PR #274, opened 2023-05-16
fix: add example values for inactivity timeout option

aj-foster:
Hello 👋🏼 

Very interesting project! The first question I ran into while trying it out was "in what format do I input the inactivity timeout?" This PR has one way of making that clearer for the docker and k8s providers.

Just food for thought. Best of luck!


FabianKramm:
@aj-foster thanks a lot for the PR! LGTM!

PR #295, opened 2023-05-17
fix: disable CGO to have statically built binaries

89luca89:
Right now Linux builds are not statically compiled

This can lead problems with systems that are too-new or too-old.
For example compiling `devpod` cli on openSUSE Tumbleweed will result in impossibility to run the agent on Ubuntu 20.40 and 22.04 containers

This PR fixes this problem

Needs to be tested on MacOS and Windows in order to be sure everything is in the right place


.github/workflows/release.yaml
```diff
@@ -422,7 +422,7 @@ jobs:
                 }
                 core.info(`Downloading ${sigAsset.name} (ID: ${sigAsset.id})`)
                 const sig = await fetchAsset(sigAsset.id)
-                
+
```
FabianKramm, line 425: this will cause issues, can you revert that?

.github/workflows/release.yaml
```diff
@@ -106,7 +106,7 @@ jobs:
         run: |
           git config --global core.autocrlf false
           git config --global core.eol lf
-          
+
```
FabianKramm, line 109: this will cause issues, can you revert that?

.github/workflows/release.yaml
```diff
@@ -263,7 +263,7 @@ jobs:
 
           $destination_path = "codesigntool"
           Write-Output "Unzipping to $destination_path"
-          Expand-Archive "codesigntool.zip" -DestinationPath $destination_path 
+          Expand-Archive "codesigntool.zip" -DestinationPath $destination_path
```
FabianKramm, line 266: please revert that as well

FabianKramm (changes requested):

PR #322, opened 2023-05-19
fix hack/rebuild.sh to not required root to have go installed

pleclech:
With the current rebuild.sh, root is required to have golang installed as ````sudo go build```` is done.
This PR allow the build to be done under the current user and then sudo is only required to move the build into destination folder


FabianKramm (approved):
@pleclech thanks for the PR! LGTM!

Issue #292, opened 2023-05-17
Using podman devpod failed to find container at creation time


pleclech:
**What happened?**
I create a workspace using the example provided but i am using podman instead of docker.
After a while when all is created podman failed with an error:

```
find container: Error: invalid argument "label=devcontainer.metadata=[{\"id\":\"ghcr.io/devcontainers/features/common-utils:2\"},{\"id\":\"./local-features/apache-config\"},{\"id\":\"ghcr.io/devcontainers/features/node:1\",\"customizations\":{\"vscode\":{\"extensions\":[\"dbaeumer.vscode-eslint\"]}}},{\"id\":\"ghcr.io/devcontainers/features/git:1\"},{\"remoteUser\":\"vscode\",\"customizations\":{\"vscode\":{\"extensions\":[\"xdebug.php-debug\",\"bmewburn.vscode-intelephense-client\",\"mrmlnc.vscode-apache\"],\"settings\":{\"php.validate.executablePath\":\"/usr/local/bin/php\"}}}},{\"customizations\":{\"vscode\":{\"extensions\":[\"streetsidesoftware.code-spell-checker\"],\"settings\":{}}}}]" for "-f, --filter" flag: parse error on line 1, column 31: bare " in non-quoted-field
See 'podman ps --help'
```

**What did you expect to happen instead?**
The intallation script to continue normally.

**How can we reproduce the bug?** (as minimally and precisely as possible)
Just create a workspace using one of the sample provided using podman as an agent
for the time being i use:

```
podman version
Client:       Podman Engine
Version:      4.5.0
API Version:  4.5.0
Go Version:   go1.20.4
Git Commit:   268511680f4a72b4a0595497a37e4d6a7da0215c
Built:        Thu May 11 11:30:51 2023
OS/Arch:      linux/amd64
```

**Local Environment:**

* DevPod Version: v0.1.2
* Operating System: windows
* ARCH of the OS: AMD64

**DevPod Provider:**

* Local/remote provider: docker | ssh
* Custom provider:  based on ssh using podman as agent instead of docker

**Anything else we need to know?**
I think the culprit is here : [find after creation](https://github.com/loft-sh/devpod/blob/b94eb4ae2afeb5e4e1e1c7e93336f89caed182dc/pkg/devcontainer/single.go#L80)
[...]

PR #323, opened 2023-05-19
fix issue 292, where find container failed when using podman

pleclech:
Fixing issue #292:
Modifying FindDevContainer to only search by the ID and not ID+metadata.

### DIFF
--- a/pkg/devcontainer/single.go
+++ b/pkg/devcontainer/single.go
@@ -77,7 +77,7 @@ func (r *Runner) runSingleContainer(parsedConfig *config.SubstitutedConfig, work
 		//TODO: wait here a bit for correct startup?
 
 		// get container details
-		containerDetails, err = r.Driver.FindDevContainer(context.TODO(), labels)
+		containerDetails, err = r.Driver.FindDevContainer(context.TODO(), labels[0:1])
 		if err != nil {
 			return nil, err
 		}


FabianKramm:
@pleclech thanks for the PR! Not sure if that is exactly what we want, as this is omitting the metadata check completely, what is the exact problem with podman and the metadata label?

pleclech:
> @pleclech thanks for the PR! Not sure if that is exactly what we want, as this is omitting the metadata check completely, what is the exact problem with podman and the metadata label?

Hello @FabianKramm ,
The problem is described in  #292, but for short podman filter failed badly with the format of the metadata and so no container can be retrieved
```
find container: Error: invalid argument "label=devcontainer.metadata=[{\"id\":\"ghcr.io/devcontainers/features/common-utils:2\"},{\"id\":\"./local-features/apache-config\"},{\"id\":\"ghcr.io/devcontainers/features/node:1\",\"customizations\":{\"vscode\":{\"extensions\":[\"dbaeumer.vscode-eslint\"]}}},{\"id\":\"ghcr.io/devcontainers/features/git:1\"},{\"remoteUser\":\"vscode\",\"customizations\":{\"vscode\":{\"extensions\":[\"xdebug.php-debug\",\"bmewburn.vscode-intelephense-client\",\"mrmlnc.vscode-apache\"],\"settings\":{\"php.validate.executablePath\":\"/usr/local/bin/php\"}}}},{\"customizations\":{\"vscode\":{\"extensions\":[\"streetsidesoftware.code-spell-checker\"],\"settings\":{}}}}]" for "-f, --filter" flag: parse error on line 1, column 31: bare " in non-quoted-field
See 'podman ps --help'
```
If you go back the second time , it will work because the first time  [you check if the container exists but only by ID](https://github.com/loft-sh/devpod/blob/b94eb4ae2afeb5e4e1e1c7e93336f89caed182dc/pkg/devcontainer/single.go#L15), and [then exit if so](https://github.com/loft-sh/devpod/blob/b94eb4ae2afeb5e4e1e1c7e93336f89caed182dc/pkg/devcontainer/single.go#L15).
I think it make sense since ID is unique and adding metadata in the filter for searching will not add any more value ?

What do you think ?

89luca89:
Hi @pleclech 
I've implemented another PR #338 which works around the label thing (which seems to be a podman upstream problem) using b64 encoding, additionally it implements some other rootless podman fixes :+1:

89luca89:
> > Hi @pleclech I've implemented another PR #338 which works around the label thing (which seems to be a podman upstream problem) using b64 encoding, additionally it implements some other rootless podman fixes +1
> 
> Hi @89luca89 I tried to pull your PR and build it, but i can't because of this error: `pkg/devcontainer/setup/setup.go:39:55: undefined: syscall.Stat_t` as `syscall.Stat_t` is Linux only

Hi @pleclech where are running podman from?
Technically that check (the stat one) should be performed inside the container, which is linux

pleclech:
> > > Hi @pleclech I've implemented another PR #338 which works around the label thing (which seems to be a podman upstream problem) using b64 encoding, additionally it implements some other rootless podman fixes +1
> > 
> > 
> > Hi @89luca89 I tried to pull your PR and build it, but i can't because of this error: `pkg/devcontainer/setup/setup.go:39:55: undefined: syscall.Stat_t` as `syscall.Stat_t` is Linux only
> 
> Hi @pleclech where are running podman from? Technically that check (the stat one) should be performed inside the container, which is linux

At compile time, and if i look in the test directory i don't see any cli for windows being built

89luca89:
I've removed that part for now, can you try again?

pleclech:
> I've removed that part for now, can you try again?

The build succeed. And podman does not complain for the filter

PR #325, opened 2023-05-19
Correctly specify how to find DevPod version

vegasje:
`version` is a command rather than a `--version` flag.


FabianKramm (approved):
@vegasje thanks a lot for the PR! LGTM!

Issue #305, opened 2023-05-17
forwardPorts not support "host:port" syntax


mrdev023:
**What happened?**

When i try to use "\[service_name\]:\[port\]" with docker-compose config. I get `json: invalid number literal, trying to unmarshal "\"db:5432\"" into Number`

See [https://containers.dev/implementors/json_reference/#general-properties](https://containers.dev/implementors/json_reference/#general-properties)

**How can we reproduce the bug?** (as minimally and precisely as possible)

My `devcontainer.json`:

```
{
    ...
    "forwardPorts": [
        3000,
        "db:5432"
    ],
    ...
}
```

**Local Environment:**

* DevPod Version: v0.1.4
* Operating System: linux
* ARCH of the OS: AMD64

**DevPod Provider:**

* Local/remote provider: docker

FabianKramm:
@mrdev023 thanks a lot for the issue! Oh true, seems like we missed that that is possible as well, should be an easy fix, we will fix that!

PR #327, opened 2023-05-19
fix: add support for "host:port" forwardPorts syntax

lizardruss:
fixes: #305
Fixes ENG-1420


pkg/devcontainer/config/config.go
```diff
@@ -354,3 +355,55 @@ func (m *Mount) UnmarshalJSON(data []byte) error {
 	}
 	return types.ErrUnsupportedType
 }
+
```
lizardruss, line 358: found StrIntArray after implementing this.. I can refactor if desired.

FabianKramm (changes requested):
@lizardruss thanks for the PR! Yeah please use types.StrIntArray instead, like we do for `AppPort`

FabianKramm (approved):

PR #332, opened 2023-05-22
replace hard written docker with workspaceInfo.Agent.Docker.Path + allow env variables to be passed into docker command

pleclech:
1. in some places docker command is hard written such as [dockerReachable](https://github.com/loft-sh/devpod/blob/7ec3edaf696fac8169e84a61a6bc7cb51e91212a/pkg/agent/agent.go\#L324), it can cause problems since the command may not exists or not the one that should be called since we explicitely provide another one

2. in some cases it will be helpfull to pass some enviromnent variables to the command, such as DOCKER_HOST, etc... So for that i added a new envs mapping into the driver docker agent section:
```
  docker:
    path: /my/custom/command
    envs:
      DOCKER_HOST: "${API_BASE_URL}"
      ...
      KEY: VALUE
    install: false
```

### DIFF
--- a/pkg/provider/provider.go
+++ b/pkg/provider/provider.go
@@ -167,7 +167,7 @@ type ProviderDockerDriverConfig struct {
 	Install types.StrBool `json:"install,omitempty"`
 
 	// Environment variables to set when running docker commands
-	Envs map[string]string `json:"envs,omitempty"`
+	Env map[string]string `json:"envs,omitempty"`
 }
 
 type ProviderAgentConfigExec struct {
--- a/pkg/provider/provider.go
+++ b/pkg/provider/provider.go
@@ -165,6 +165,9 @@ type ProviderDockerDriverConfig struct {
 
 	// If false, DevPod will not try to install docker into the machine.
 	Install types.StrBool `json:"install,omitempty"`
+
+	// Environment variables to set when running docker commands
+	Envs map[string]string `json:"envs,omitempty"`
 }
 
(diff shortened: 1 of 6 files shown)


pkg/provider/provider.go
```diff
@@ -165,6 +165,9 @@ type ProviderDockerDriverConfig struct {
 
 	// If false, DevPod will not try to install docker into the machine.
 	Install types.StrBool `json:"install,omitempty"`
+
+	// Environment variables to set when running docker commands
+	Envs map[string]string `json:"envs,omitempty"`
```
FabianKramm, line 170: Can we rename that to `env`?

FabianKramm (changes requested):
@pleclech thanks a lot for the PR! Almost looks good, just want a small change then we can merge it!

pkg/provider/provider.go
pleclech: Hi @FabianKramm
Of course no problems.

pleclech: @FabianKramm Done.

FabianKramm (approved):

pleclech:
@FabianKramm I forget to rename the ```json:"envs,omitempty"``` to ````json:"env,omitempty"``` : [envs metadata for the json field](https://github.com/loft-sh/devpod/blob/d2d76ca66ec486958c830d766d1e449ca2113408/pkg/provider/provider.go#L170) !

PR #337, opened 2023-05-22
Pass runArgs from devcontainer.json into docker run command

vegasje:
`runArgs` set in `devcontainer.json` are currently parsed by DevPod, but they are not added when the `docker run` command is issued.

This change passes `runArgs` into the `docker run` command.

--------------------------------

I found the need for `runArgs` as I was attempting to set up multiple containers that could communicate with each other.  Docker Compose is often used for this scenario, but I needed to work with Docker networking directly.

Being able to pass the `runArgs` to the `docker run` command allowed me to achieve this:

```
{
	"name": "devpod-test",
	"build": {
		"dockerfile": "Dockerfile"
	},
	"initializeCommand": "docker network inspect test-network >/dev/null 2>&1 || docker network create --driver bridge test-network",
	"runArgs": [
		"--name=devpod-test",
		"--network=test-network"
	],
	...
}
```

### DIFF
--- a/pkg/driver/docker/docker.go
+++ b/pkg/driver/docker/docker.go
@@ -197,6 +197,9 @@ func (d *dockerDriver) RunDevContainer(
 		}
 	}
 
+	// runArgs
+	args = append(args, parsedConfig.RunArgs...)
+
 	// run detached
 	args = append(args, "-d")
 


davemaple:
+1 

we will utilize this PR once adopted as well.

adds support for **_Image or Dockerfile-specific properties_** of `devcontainer.json`, which seems to align with the project goal of supporting:
https://containers.dev/implementors/json_reference/#image-specific

PR #338, opened 2023-05-23
fix: fix podman support in docker driver

### LINKED ISSUE #292: Using podman devpod failed to find container at creation time

89luca89:
In this PR we fix:

- a special case with complex map labels that leads to #292 
- rootless podman support
- user mapping related issues while chowning home and workspace

Fix #292 
Resolves ENG-1411


pkg/driver/docker/docker.go
```diff
@@ -152,6 +153,16 @@ func (d *dockerDriver) RunDevContainer(
 	if mergedConfig.Privileged != nil && *mergedConfig.Privileged {
 		args = append(args, "--privileged")
 	}
+
+	// In case we're using podman, let's use userns to keep
+	// the ID of the user (vscode) inside the container as
+	// the same of the external user.
+	// This will avoid problems of mismatching chowns on the
+	// project files.
+	if d.Docker.DockerCommand == "podman" {
```
FabianKramm, line 162: What happens if the user uses podman as an alias to docker? Is there a way we can find out if the docker command itself is podman?

pkg/docker/helper.go
```diff
@@ -157,12 +158,51 @@ func (r *DockerHelper) Inspect(ids []string, inspectType string, obj interface{}
...
 	for _, label := range labels {
 		args = append(args, "--filter", "label="+label)
 	}
 
+	out, err := r.buildCmd(context.TODO(), args...).Output()
+	if err != nil {
+		labelsB64 := []string{}
+		for _, label := range labels {
+			labelb64 := base64.StdEncoding.EncodeToString([]byte(label))
+			labelsB64 = append(labelsB64, labelb64)
```
FabianKramm, line 177: Rewrite as:
```
labelsB64 = append(labelsB64, base64.StdEncoding.EncodeToString([]byte(label)))
```

Also please move the translation of the labels into the FindContainerB64 function

FabianKramm (changes requested):

pkg/docker/helper.go
```diff
@@ -182,6 +191,42 @@ func (r *DockerHelper) FindContainer(labels []string) ([]string, error) {
 	return arr, nil
 }
 
+// FindContainerJSON will manually search for containers with matching labels.
+// This is useful in case the `--filter` doesn't work.
+func (r *DockerHelper) FindContainerJSON(labels []string) ([]string, error) {
+	args := []string{"ps", "-q", "-a"}
```
FabianKramm, line 197: Can we not use --format here to run a single command to find all container labels?

FabianKramm (approved):

PR #346, opened 2023-05-25
fix: add sudo password check

inhumantsar:
When connecting to an SSH provider with a user account that needs to enter a password when running `sudo`, the default behaviour is to fail silently and retry until timing out. Debug output (see below) makes the problem obvious, but it would still be better if the injection process failed early with a clear message.

Ideally, there would a way to flag up the need for a password to the desktop user and ask them to enter it, but I'm hoping this will do as a stop gap for now.

Open to guidance on how to handle this more gracefully.

```
[18:30:50] info Waiting for devpod agent to come up...
[18:30:50] debug Inject Error: bash: cannot set terminal process group (4333): Inappropriate ioctl for device
bash: no job control in this shell
sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
sudo: a password is required
EOF
```

### DIFF
--- a/pkg/inject/inject.sh
+++ b/pkg/inject/inject.sh
@@ -26,6 +26,9 @@ command_exists() {
   command -v "$@" >/dev/null 2>&1
 }
 
+sudo_password_required() {
+}
+
 is_arm() {
   case "$(uname -a)" in
   *arm* ) true;;
@@ -78,6 +81,12 @@ if {{ .ExistsCheck }}; then
   sh_c='sh -c'
   if [ "$user" != 'root' ]; then
     if command_exists sudo; then
+      # check if sudo requires a password
+      if echo $(sudo -n uptime 2>&1) | grep -q "a password is required"; then
+        >&2 echo Error: sudo requires a password and no password is available. Please ensure your user account is configured with NOPASSWD.
+        exit 1
+      fi
+
       sh_c='sudo -E sh -c'
     elif command_exists su; then
       sh_c='su -c'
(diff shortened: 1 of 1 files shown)


pkg/inject/inject.sh
```diff
@@ -78,6 +78,12 @@ if {{ .ExistsCheck }}; then
   sh_c='sh -c'
   if [ "$user" != 'root' ]; then
     if command_exists sudo; then
+      # check if sudo requires a password
+      if echo $(sudo -n uptime 2>&1) | grep -q "a password is required"; then
```
89luca89, line 82: I think here we can simply check the return code:

`if ! sudo -n uptime; then`

Should work the same way

89luca89 (changes requested):

pkg/inject/inject.sh
FabianKramm: I think we need to make sure the command doesn't output anything on stdout or stderr as otherwise the communication could get broken

89luca89: Yes you can redirect both strerr and out to dev/null, but the thing to check is more the return code than the output (that might change in some ways in the future)

dgl: Also rather than using `uptime`, doing `sudo -nl` (list available commands, see the "EXIT VALUE" section of man sudo) is cleaner (it avoids a log about the use of sudo, as it's clearly a no-op command). Leading to:

```shell
if ! sudo -nl >/dev/null 2>&1; then
```

inhumantsar: thanks, that's tidy. using that

89luca89 (approved):
LGTM

PR #357, opened 2023-05-28
fix: update add provider from url example to use a working url

inhumantsar:
The SSH provider's YAML file is actually a template which is prepared for release by GitHub Actions. Targeting it with `devpod provider add` results in an error, users should target the GitHub release download instead.


FabianKramm (approved):
@inhumantsar thanks a lot for the PR! LGTM!

PR #358, opened 2023-05-29
fix: resolve openvscode process eating CPU when using docker

89luca89:
When using Openvscode (Web Vscode) with Docker, a process keeps 1 core of the cpu binned to 100%.
This does not happen in Podman.

Turns out this is because of a permission problem, where the `.openvscode` directory had some inaccessible files/dirs for permissions.

This PR fixes this, and contextually upgrades openvscode to the latest release v1.78.2


pascalbreuninger (approved):
lgtm

ThomasK33 (approved):

PR #359, opened 2023-05-29
fix panic : fatal agent error: panic: runtime error: index out of range [2]

pleclech:
I just fall on this bug today with  fatal agent error: panic: runtime error: index out of range [2] with length 2 goroutine 1.

And after investigating found that [i is incremented when it should be decremented](https://github.com/loft-sh/devpod/blob/204d0decc41768ff5494c396462220c948d14995/pkg/dockerfile/parse.go#L141)  also [here](https://github.com/loft-sh/devpod/blob/204d0decc41768ff5494c396462220c948d14995/pkg/dockerfile/parse.go#L127)


89luca89 (approved):
LGTM

PR #360, opened 2023-05-29
fix: implement volume deletion when removing a devcontainer

89luca89:
Right now deleting a workspace will leave its docker volumes behing
This implements the volume deletion when removing a devcontainer

### DIFF
--- a/pkg/driver/docker/docker.go
+++ b/pkg/driver/docker/docker.go
@@ -101,7 +101,6 @@ func (d *dockerDriver) DeleteDevContainer(ctx context.Context, id string, delete
 		volume = container[0].Config.Labels["dev.containers.id"]
 	}
 
-
 	err := d.Docker.Remove(ctx, id)
 	if err != nil {
 		return err
--- a/pkg/driver/docker/docker.go
+++ b/pkg/driver/docker/docker.go
@@ -98,6 +98,9 @@ func (d *dockerDriver) DeleteDevContainer(ctx context.Context, id string, delete
 		if err != nil {
 			return err
 		}
+		if len(container) == 0 {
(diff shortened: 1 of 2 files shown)


pascalbreuninger (approved):

pkg/driver/docker/docker.go
```diff
@@ -90,9 +90,27 @@ func (d *dockerDriver) StartDevContainer(ctx context.Context, id string, labels
...
+	var volume string
 
-	return d.Docker.Remove(ctx, id)
+	if deleteVolumes {
+		// inspect container deleteVolumes
+		container, err := d.Docker.InspectContainers([]string{id})
+		if err != nil {
+			return err
+		}
+		volume = container[0].Config.Labels["dev.containers.id"]
```
ThomasK33, line 101: Is there a chance that Labels might be nil?

Alternatively, what if docker daemon misbehaves for some reason and len(containers) = 0?

pkg/driver/docker/docker.go
```diff
@@ -90,9 +90,27 @@ func (d *dockerDriver) StartDevContainer(ctx context.Context, id string, labels
...
+	}
+
+	err := d.Docker.Remove(ctx, id)
+	if err != nil {
```
ThomasK33, line 105: Personal preference, so feel free to ignore.

```suggestion
	if err := d.Docker.Remove(ctx, id); err != nil {
```

pkg/driver/docker/docker.go
89luca89: Done :+1: Thanks!

ThomasK33 (approved):

PR #378, opened 2023-06-08
fix: do not use sudo, if it's not strictly needed.

### LINKED ISSUE #306: Hangs trying to configure any provider
**What happened?**
Trying to add any provider hangs on "Loading <provider name>..."

Anything

AppImage on Debian Bookworm

**Local Environment:**

* DevPod Version: 0.1.5
* Operating System: linux
* ARCH of the OS: AMD64

Bottom of DevPod window also has incorrect system information. "Version 0.1.5 | unknown platform | unknown arch"

### LINKED ISSUE #354: Can't create a workspace using SSH provider due to a tty issue
Disclaimer: I am not entirely sure whether it's a bug or just me missing some detail 😅 

**What happened?**  
I am trying to set up an remote workspace using the SSH provider. When creating the workspace (I'm using the `--debug` flag) I keep repeatedly seeing the following error:

```
[11:00:35] info Creating devcontainer...
[11:00:35] debug Inject and run command: '/home/<redacted>/devpod/agent' agent workspace up --workspace-info '<long-token>' --debug
[11:00:35] debug execute inject script
[11:00:35] debug Run command provider command: ssh -oStrictHostKeyChecking=no \
    -p ${PORT} \
    ${EXTRA_FLAGS} \
    "${HOST}" \
    "${COMMAND}"
[11:00:35] debug Pseudo-terminal will not be allocated because stdin is not a terminal.
[11:00:35] debug sudo: sorry, you must have a tty to run sudo
[11:00:35] debug done exec
[11:00:35] debug done inject
[11:00:35] debug done injecting
[11:00:35] debug Inject Error: Pseudo-terminal will not be allocated because stdin is not a terminal.
sudo: sorry, you must have a tty to run sudo
EOF
```
I expected the workspace to be created successfully.

My *`devcontainer.json`*:
```
{
    "name": "Test",
    "image": "quay.io/qiime2/core:2023.2",
    "forwardPorts": []
}
```

The SSH provider configuration:
All default settings except agent path (set to `/home/<redacted>/devpod/agent`, even though the same behaviour was observed with the default value).
I also tried to add the `-t`SSH flag, did not make a difference.

**Local Environment:**  
- DevPod Version: v0.1.7
- Operating System: mac
- ARCH of the OS: ARM64

**DevPod Provider:**  
- _remote_ provider: ssh

I am using zsh on the remote.

89luca89:
With this, we go for the `sudo` route only if we can't actually write in the INSTALL_PATH
This way (and with the updated ssh provider defaults in the next release) We can work on remote machines that do not have sudo rights.

Resolves #354 
Resolves #306 
Resolves ENG-1421
Resolves ENG-1480

### DIFF
--- a/pkg/inject/inject.sh
+++ b/pkg/inject/inject.sh
@@ -76,14 +76,17 @@ download() {
 if {{ .ExistsCheck }}; then
   user="$(id -un || true)"
   sh_c='sh -c'
-  if [ "$user" != 'root' ]; then
+
+  # Try to create the install dir, if we fail, we search for sudo
+  # else let's continue without sudo, we don't need it.
+  $sh_c "mkdir -p $INSTALL_DIR"
+  if [ "$?" -ne 0 ]; then
     if command_exists sudo; then
       # check if sudo requires a password
       if ! sudo -nl >/dev/null 2>&1; then
         >&2 echo Error: sudo requires a password and no password is available. Please ensure your user account is configured with NOPASSWD.
         exit 1
       fi
-
       sh_c='sudo -E sh -c'
     elif command_exists su; then
       sh_c='su -c'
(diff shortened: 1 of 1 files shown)


pascalbreuninger (approved):
LGTM

PR #380, opened 2023-06-09
fix(#163): handle objects in the devcontainer lifecyle hooks

### LINKED ISSUE #163: Lifecycle script object format and parallel execution
Currently lifecycle scripts only support the array specification.

Object based lifecycle scripts are executed in parallel, but are not currently supported.

Please see the [parallel lifecycle script execution](https://containers.dev/implementors/spec/#parallel-exec) section and [lifecycle scripts](https://containers.dev/implementors/json_reference/#lifecycle-scripts) in the devcontainer specification

mrsimonemms:
The [lifestyle hooks](https://containers.dev/implementors/spec/#parallel-exec) in the devcontainer spec accept an object of key/value pairs as well as strings and array of strings.

This PR adds support for that. Internally, it converts it to an array of strings (as the strings are as well) and, from a Go point of view, it creates a new `type.StrObjectArray`.

This is my first contribution to this project, so please let me know if there's anything I need to do additionally to get this accepted.

Fixes #163

### DIFF
--- a/cmd/provider/list-default-providers.go
+++ b/cmd/provider/list-default-providers.go
@@ -1,0 +1,79 @@
+package provider
+
+import (
+	"context"
+	"crypto/tls"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/loft-sh/devpod/cmd/flags"
+	"github.com/spf13/cobra"
+)
+
+// ListAvailableCmd holds the list cmd flags
+type ListAvailableCmd struct {
+	flags.GlobalFlags
+}
+
+var httpClient = &http.Client{
+	Transport: &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+	},
+}
+
+func getDevpodProviderList() error {
+	req, err := http.NewRequest("GET", "https://api.github.com/users/loft-sh/repos", nil)
+	if err != nil {
+		return err
+	}
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		return err
+	}
+
+	result, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+
+	var jsonResult []map[string]interface{}
+	err = json.Unmarshal(result, &jsonResult)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("List of available providers from loft:")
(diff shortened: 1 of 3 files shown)


pascalbreuninger:
@mrsimonemms  Thank you for opening this PR 🙇 

There are two issues I currently see with this approach:
1. How do we concatenate multiple object commands?
2. Per spec object commands should be executed in parallel

Now I'd just leave the second concern aside for now, can implement this later.
The first one however will cause issues. Given the `postStartCommand` of 
```json
{ "foo": "echo foo; sleep 60", "bar": "echo bar" }
``` 
the current implementation serializes this to an array command: `echo foo; sleep 60 echo bar` without delimiter between commands

pascalbreuninger (changes requested):
@mrsimonemms  Thank you for opening this PR 🙇 

There are two issues I currently see with this approach:
1. How do we concatenate multiple object commands?
2. Per spec object commands should be executed in parallel

Now I'd just leave the second concern aside for now, can implement this later.
The first one however will cause issues. Given the `postStartCommand` of 
```json
{ "foo": "echo foo; sleep 60", "bar": "echo bar" }
``` 
the current implementation serializes this to an array command: `echo foo; sleep 60 echo bar` without delimiter between commands

mrsimonemms:
Hi @pascalbreuninger

I think you're right about leaving the second part for the moment. I've treated this as an MVP, (ie, parsing the incoming JSON into DevPod format) so the _how_ can be implemented later as it would make this change much bigger.

I'm a bit confused on the other part. How/why does it concatenate `{ "foo": "echo foo; sleep 60", "bar": "echo bar" }` to `echo foo; sleep 60 echo bar`? This converts the input from an object of strings to an array of strings.

I have just noticed one thing it's missing though - it only covers objects of strings, but the spec also has objects of arrays of strings which I will implement this evening.

pascalbreuninger:
yep, but as far as I can tell, separate commands aren't delimited in any way but just seperated by spaces, resulting in broken commands

mrsimonemms:
Hmmm, maybe I've misunderstood it then - I've done the `StrObjectArray` as an exact copy of the `StrArray` with the additional object handler. My assumption was that the array of strings was full commands, not one chunk of the command in each string (admittedly, this would make sense as the Docker/Kubernetes way of doing the commands).

Let me have a think how it could work.

PR #388, opened 2023-06-15
feat: add provider command to list installable providers

89luca89:
This should ease the life of a CLI-only user.

It gets on par with what is already available on GUI, a list of our 1st party providers ready to install.

Resolves ENG-1326

### DIFF
--- a/cmd/provider/list-default-providers.go
+++ b/cmd/provider/list-default-providers.go
@@ -16,9 +16,6 @@ import (
 // ListAvailableCmd holds the list cmd flags
 type ListAvailableCmd struct {
 	flags.GlobalFlags
-
-	Output string
-	Used   bool
 }
 
 var httpClient = &http.Client{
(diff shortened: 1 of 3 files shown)


cmd/provider/list-default-providers.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("List of available providers from loft:")
+	for _, v := range jsonResult {
+		if strings.Contains(v["name"].(string), "devpod-provider") {
+			name := strings.Split(v["name"].(string), "-")
+			fmt.Println("\t", name[len(name)-1])
+		}
```
ThomasK33, line 56: What if there will be a provider called "devpod-provider-[cloud]-v2" in the future?

Might be worth _only_ stripping "devpod-provider-" from the name and using the rest.

89luca89: makes sense yes :+1:

cmd/provider/list-default-providers.go
```diff
@@ -0,0 +1,82 @@
...
+
+	"github.com/loft-sh/devpod/cmd/flags"
+	"github.com/spf13/cobra"
+)
+
+// ListAvailableCmd holds the list cmd flags
+type ListAvailableCmd struct {
+	flags.GlobalFlags
+
+	Output string
```
pascalbreuninger, line 20: Do we need these? (`output` and `used`)

pascalbreuninger (approved):

PR #389, opened 2023-06-16
Avoid "Text file busy" during download/injection

anguslees:
Don't download/inject to the same inode, in case it is already in use.

This PR takes the conservative approach of writing to a tempfile and then
renaming over the original in a second step.  This means a later (failed)
download/inject won't corrupt any existing file.

Fixes "Text file busy" error:

```
16:57:57 info Inject binary
16:57:57 info done exec
16:57:58 info Done injecting binary
16:57:58 info done inject
16:57:58 info done injecting
16:57:58 info Inject Error: sh: 1: cannot create /usr/local/bin/devpod: Text file busy
16:57:58 info write |1: file already closed
```

### DIFF
--- a/pkg/inject/inject.sh
+++ b/pkg/inject/inject.sh
@@ -38,7 +38,8 @@ is_arm() {
 
 inject() {
   echo "ARM-$(is_arm && echo -n 'true' || echo -n 'false')"
-  $sh_c "cat > $INSTALL_PATH"
+  $sh_c "cat > $INSTALL_PATH.$$"
+  $sh_c "mv $INSTALL_PATH.$$ $INSTALL_PATH"
 
   if [ "$CHMOD_PATH" = "true" ]; then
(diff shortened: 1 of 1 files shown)


FabianKramm (changes requested):
@anguslees thanks a lot for the PR! Good catch, but I guess we need 2 changes:
1. Move it to before download and inject to always execute as otherwise the same issue can occur when downloading and writing to the install path
2. Make sure to ignore when `rm` fails and pipe stderr to /dev/null

anguslees:
Done.  I modified the approach to write to a tempfile and _then_ rename into place, so it will be robust to failures on the second download.   This might be over-engineering, since I don't think there's a case where the existing binary is re-executed - but I don't fully understand how/where all this gets invoked yet 😅.

FabianKramm (approved):

FabianKramm:
@anguslees looks good now, thanks a lot!

PR #411, opened 2023-06-21
fix: handle rootful podman

89luca89:
This PR will fix support for rootful (non-rootless) podman

userns keep-id is reserved only for rootless podman, so let's check if we're root before adding the flag


ThomasK33 (approved):
LGTM

PR #422, opened 2023-06-21
fix: properly check kubectl success on init. Force workspace folder deletion

89luca89:
Right now the k8s init wasn't working properly, allowing for invalid kubeconfig paths, and values
Contextually this would lead to `delete` to not actually delete the workspace dir, thus making the user stuck with the "wrong" config, not able to update it and requiring manual intervention.

Resolves ENG-1600

### DIFF
--- a/cmd/delete.go
+++ b/cmd/delete.go
@@ -112,7 +112,7 @@ func (cmd *DeleteCmd) Run(ctx context.Context, devPodConfig *config.Config, args
 		GracePeriod: duration,
 	})
 	if err != nil {
-		return errors.Wrap(err, clientimplementation.DeleteWorkspaceFolder(client.Context(), client.Workspace(), log.Default).Error())
+		return err
 	}
 
 	log.Default.Donef("Successfully deleted workspace '%s'", client.Workspace())
--- a/cmd/delete.go
+++ b/cmd/delete.go
@@ -112,10 +112,7 @@ func (cmd *DeleteCmd) Run(ctx context.Context, devPodConfig *config.Config, args
 		GracePeriod: duration,
 	})
 	if err != nil {
-		err2 := clientimplementation.DeleteWorkspaceFolder(client.Context(), client.Workspace(), log.Default)
-		if err2 != nil {
-			return errors.Wrap(err2, err.Error())
-		}
+		return errors.Wrap(err, clientimplementation.DeleteWorkspaceFolder(client.Context(), client.Workspace(), log.Default).Error())
 	}
 
 	log.Default.Donef("Successfully deleted workspace '%s'", client.Workspace())
--- a/cmd/delete.go
+++ b/cmd/delete.go
@@ -112,7 +112,10 @@ func (cmd *DeleteCmd) Run(ctx context.Context, devPodConfig *config.Config, args
 		GracePeriod: duration,
 	})
 	if err != nil {
-		return err
+		err2 := clientimplementation.DeleteWorkspaceFolder(client.Context(), client.Workspace(), log.Default)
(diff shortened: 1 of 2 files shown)


cmd/delete.go
```diff
@@ -112,7 +112,10 @@ func (cmd *DeleteCmd) Run(ctx context.Context, devPodConfig *config.Config, args
 		GracePeriod: duration,
 	})
 	if err != nil {
-		return err
+		err2 := clientimplementation.DeleteWorkspaceFolder(client.Context(), client.Workspace(), log.Default)
+		if err2 != nil {
+			return errors.Wrap(err2, err.Error())
+		}
```
ThomasK33, line 118: How about:

```suggestion
		return errors.Join(err, clientimplementation.DeleteWorkspaceFolder(client.Context(), client.Workspace(), log.Default))
```

At the end of the day we've encountered an error and should probably return, independent of the workspace delete failing or not.

89luca89: Done

ThomasK33 (approved):
LGTM

cmd/delete.go
```diff
@@ -112,7 +112,7 @@ func (cmd *DeleteCmd) Run(ctx context.Context, devPodConfig *config.Config, args
 		GracePeriod: duration,
 	})
 	if err != nil {
-		return err
+		return errors.Wrap(err, clientimplementation.DeleteWorkspaceFolder(client.Context(), client.Workspace(), log.Default).Error())
```
FabianKramm, line 115: I'm not sure this is a good idea, we have the `--force` flag to make sure even if the deletion fails the workspace is deleted.

FabianKramm (changes requested):

cmd/delete.go
89luca89: This was not working with the `--force` flag, even with it, it was stopping at this very point

providers/kubernetes/provider.yaml
```diff
@@ -1,5 +1,5 @@
 name: kubernetes
-version: v0.0.1
+version: v0.0.2
```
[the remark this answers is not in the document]
89luca89: ah ok I thought it was needed for the update

providers/kubernetes/provider.yaml
```diff
@@ -95,11 +95,11 @@ agent:
 exec:
   init: |-
     # Check if kubernetes is reachable
-    KUBERNETES_REACHABLE=$(KUBECONFIG=${KUBERNETES_CONFIG} ${KUBECTL_PATH} get pods --namespace=${KUBERNETES_NAMESPACE} --context=${KUBERNETES_CONTEXT} > /dev/null 2>&1 && echo -n "true" || echo -n "false")
-    if [ "KUBERNETES_REACHABLE" = "false" ]; then
+    KUBERNETES_REACHABLE=$(${KUBECTL_PATH} get pods --kubeconfig=${KUBERNETES_CONFIG} --namespace=${KUBERNETES_NAMESPACE} --context=${KUBERNETES_CONTEXT} > /dev/null 2>&1 && echo -n "true" || echo -n "false")
```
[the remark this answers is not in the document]
89luca89: Using `--kubeconfig=` is the method we use inside the provider commands, and in the driver
This for example makes `~/.kube/config` behave differently between `init` and the other commands, as here, the `KUBECONFIG=${KUBERNETES_CONFIG}` will expand the `~` while `--kubeconfig=` won't, and will fail afterwards

This unifies the behaviour between the init and the rest of the commands

FabianKramm (approved):

PR #453, opened 2023-06-29
Restore http  features like proxy handling

### LINKED ISSUE #365: HTTP proxy

Yes: I'm unable to use DevPod in corporate environment.

Add support for HTTP proxy, via settings or (in Linux) by supporting standard variables like http_proxy, no_proxy and/or upcase versions.

**Additional context**

The error occured when trying to download [https://github.com/gitpod-io/openvscode-server/releases/download/openvscode-server-v1.76.2/openvscode-server-v1.76.2-linux-x64.tar.gz](https://github.com/gitpod-io/openvscode-server/releases/download/openvscode-server-v1.76.2/openvscode-server-v1.76.2-linux-x64.tar.gz)

gbonnefille:
Fixes #365

### DIFF
--- a/cmd/agent/docker_credentials.go
+++ b/cmd/agent/docker_credentials.go
@@ -6,13 +6,13 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"net/http"
 	"os"
 	"strconv"
 	"strings"
 
 	"github.com/loft-sh/devpod/cmd/flags"
 	"github.com/loft-sh/devpod/pkg/dockercredentials"
+	devpodhttp "github.com/loft-sh/devpod/pkg/http"
 	"github.com/loft-sh/log"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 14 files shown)


gbonnefille:
Unable to test on my side. Any feedback from other user is welcome.

gbonnefille:
Humh... I understood that there are many proxy related problems.

The client itself should be able to pass the proxy. I think this is solved by this PR.

But the agent is also responsible of downloading the openvscode archive. Thus, the whole Container should be configured with HTTP proxy settings.
This can be achieve by adding proxy settings in https://github.com/loft-sh/devpod/blob/43eac112842697be4882269bf0b7a7866401890a/pkg/driver/kubernetes/run.go#L138

But we have one major issue: we can imagine that the client need a proxy (because inside the corporate network) and the kubernetes cluster does not requires, as it is running outside (managed kubernetes? AWS, Google...?)

One solution can be to add settings in the provider config directly: this way, one can adjust the settings to the effective situation.

Another solution can be to change the architecture: the openvscode (and other) are no more downloaded by the agent, but by the client itself and then, injected into the container. This can be much more longer, but we do not have to configure proxy settings onto the container.

@89luca89 any thought?

FabianKramm:
@gbonnefille thanks a lot for your PR! Looks good to me, but since we essentially reuse the same piece of code multiple times there, would be good to create a function and then just call that so we have it all in a central place

FabianKramm:
Just put a function like this into package `pkg/http`:
```
func GetClient() *http.Client {
	customTransport := http.DefaultTransport.(*http.Transport).Clone()
	customTransport.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
	return &http.Client{Transport: customTransport}
}
```

FabianKramm (approved):

gbonnefille:
@FabianKramm I fear only half of the problem is solved.
As pointed previously, the agent is also doing HTTP requests. 

Should we create a new issue for that?

PR #466, opened 2023-07-03
tests: add integration tests for SSH provider

89luca89:
This will add a new tests that runs the integration with the ssh-provider
This goes in hand with https://github.com/loft-sh/devpod-provider-ssh/pull/6 in order to test
compatibility both on devpod changes and provider changes.

### DIFF
--- a/e2e/tests/integration/integration.go
+++ b/e2e/tests/integration/integration.go
@@ -11,7 +11,7 @@ import (
 
 var _ = ginkgo.Describe("[integration]: devpod provider ssh test suite", ginkgo.Ordered, func() {
 
-	ginkgo.FIt("should add provider to devpod", func() {
+	ginkgo.It("should add provider to devpod", func() {
 		// ensure we don't have the ssh provider present
 		cmd := exec.Command("bin/devpod-linux-amd64", "provider", "delete", "ssh")
 		err := cmd.Run()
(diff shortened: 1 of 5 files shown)


pascalbreuninger:
@89luca89 I'm not sure about the current state of the integration test setup. Do we wipe the `DEVPOD_HOME` directory between test suites?

pascalbreuninger (commented):
@89luca89 I'm not sure about the current state of the integration test setup. Do we wipe the DEVPOD_HOME directory between test suites?

e2e/tests/integration/integration.go
```diff
@@ -0,0 +1,84 @@
...
+
+			defer f.Close()
+			_, err = f.Write(publicKey)
+			framework.ExpectNoError(err)
+		}
+	})
+
+	ginkgo.It("should add provider to devpod", func() {
+		// ensure we don't have the ssh provider present
+		cmd := exec.Command("bin/devpod-linux-amd64", "provider", "delete", "ssh")
```
FabianKramm, line 51: Don't hardcode the binary path here, we have a function for this as otherwise this test doesn't work locally for ARM computers

FabianKramm: Please use the `f.DevPodProviderDelete` function instead

FabianKramm (changes requested):

FabianKramm:
@89luca89 please exchange your custom `exec.Command` with the correct framework functions for those commands

e2e/tests/integration/integration.go
89luca89: Done

FabianKramm (approved):

PR #468, opened 2023-07-03
fix: make setopt command not fail

### LINKED ISSUE #464: Inject Error: sh: 5: setopt: not found
When trying to start a workspace using the Docker provider, I'm seeing an infinite loop of this:

```
Waiting for devpod agent to come up...
Inject Error: sh: 5: setopt: not found
```

It seems to be [this line](https://github.com/loft-sh/devpod/blob/ddae80c2dbd3e067c4cc995c5403582f7fabd42a/pkg/inject/inject.sh#L5).

* As I'm on macOS, my shell is of course `zsh`.
* `setopt` is available in `zsh`, but it's not found if I try to run it using `/bin/sh`. The error is the same as above.
* I notice that the shebang in `inject.sh` specifies `/bin/sh` - so presumably this script is being executed using `sh`, which doesn't have `setopt` on my macOS machine.

**Local Environment:**

* DevPod Version: 0.1.10
* Operating System: macOS Ventura 13.4
* ARCH of the OS: ARM64 (M1 Max, Macbook Pro)

**DevPod Provider:**

* Local/remote provider: docker

89luca89:
This will

Fix #464 
Resolve ENG-1676

Setop is needed on ZSH if it's directly called as such, so let's allow it to fail in case

### DIFF
--- a/pkg/inject/inject.sh
+++ b/pkg/inject/inject.sh
@@ -1,9 +1,8 @@
 #!/bin/sh
 set -e
 
-if [ "$SHELL" != "${SHELL%"/zsh"*}" ]; then
-  setopt SH_WORD_SPLIT || :
-fi
+# Compatibility fix for zsh shells
+setopt SH_WORD_SPLIT 2>/dev/null || :
 
 INSTALL_DIR="{{ .InstallDir }}"
 INSTALL_FILENAME="{{ .InstallFilename }}"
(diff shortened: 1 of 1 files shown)


pascalbreuninger (approved):

liamdon:
Wow, thank you for such a speedy fix and review @89luca89 @pascalbreuninger!

PR #469, opened 2023-07-03
fix: add getent based login shell discovery

89luca89:
This will use the libc utility `getent` to infere the user's login shell

Fallback to the old method if unsuccessful

Resolves ENG-1657

### DIFF
--- a/pkg/ssh/server/ssh.go
+++ b/pkg/ssh/server/ssh.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"os/user"
+	"path/filepath"
 	"runtime"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


pkg/ssh/server/ssh.go
```diff
@@ -99,7 +100,20 @@ type Server struct {
 func getShell() ([]string, error) {
 	// try to get a shell
 	if runtime.GOOS != "windows" {
-		_, err := exec.LookPath("bash")
+		// infere login shell from getent
+		cmd := exec.Command("getent", "passwd", os.Getenv("USER"))
+		output, err := cmd.Output()
```
FabianKramm, line 105: Please create a function for this and use it like this:
```
func GetUserShell() (string, error) {
    currentUser, err := user.Current()
    if err != nil {
      return err
    }

    output, err := exec.Command("getent", "passwd", currentUser.Name).Output()
    ...
}

```

FabianKramm (changes requested):

pkg/ssh/server/ssh.go
89luca89: done

FabianKramm (approved):

PR #474, opened 2023-07-04
Support workspaces from branches with slash in the name.

levsha:
DevPod supports already creating workspaces from non-default branch by providing repository URL in the form of
`repository.host.abc/user/repo.git@branch`

With this change it accepts branches that contain slash in the name as this is common to name branch like `username/some-description`


pascalbreuninger (approved):
LGTM

PR #485, opened 2023-07-06
fix(#163): handle objects in the devcontainer lifecyle hooks

### LINKED ISSUE #163: Lifecycle script object format and parallel execution

mrsimonemms:
Following on from #380, this is a second attempt at solving #163 thanks to @pascalbreuninger's previous review.

Fixes #163 

This converts objects to an array of strings using a parallel Bash command. This does assume that `bash` is present on the devcontainers, which is probably a safe bet but I guess we could switch to `sh` if necessary.

### DIFF
--- a/examples/object-lifecycle-hooks/.devcontainer.json
+++ b/examples/object-lifecycle-hooks/.devcontainer.json
@@ -1,0 +1,9 @@
+{
+  "name": "Object-Lifecycle",
+  "image": "mcr.microsoft.com/vscode/devcontainers/go:1",
+  "postStartCommand": {
+    "cobra-install": "go install github.com/spf13/cobra-cli@latest",
+    "golangci-lint": "go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest",
+    "errcheck": "go install github.com/kisielk/errcheck@latest"
+  }
+}
--- a/pkg/devcontainer/config/config.go
+++ b/pkg/devcontainer/config/config.go
@@ -68,7 +68,7 @@ type DevContainerConfigBase struct {
 	RemoteUser string `json:"remoteUser,omitempty"`
 
 	// A command to run locally before anything else. This command is run before "onCreateCommand". If this is a single string, it will be run in a shell. If this is an array of strings, it will be run as a single command without shell.
-	InitializeCommand types.StrArray `json:"initializeCommand,omitempty"`
+	InitializeCommand types.LifecycleHook `json:"initializeCommand,omitempty"`
 
 	// Action to take when the user disconnects from the container in their editor. The default is to stop the container.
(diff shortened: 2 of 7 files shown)


pascalbreuninger:
@mrsimonemms Thanks for giving this another shot! I don't feel good about bending the existing implementation to accomodate for the object based commands. Imo we should rather refactor the way we currently parse the commands so we're spec compliant.  

If you want to go ahead and do that, I'll do my best to support you. If not I'll put it into our roadmap and will work on it rather sooner than later

mrsimonemms:
@pascalbreuninger yeah, I'm sort of with you on that. This is probably "good enough" for testing/known repos, but not for GA.

I'm happy to have a bash at it properly - can you give me a clue as to whereabouts in the code base I should be looking for the lifecycle commands and the implementation?

FabianKramm:
@mrsimonemms you can find that at https://github.com/loft-sh/devpod/blob/main/pkg/devcontainer/setup/setup.go#L232, I guess we can still execute these sequentially in the beginning if thats easier to implement and then eventually do the proper execute at the same time implementation.

mrsimonemms:
@FabianKramm thanks. I was looking in the right place then 🎉 

As an aside, there's no `initializeCommand` which is [documented](https://containers.dev/implementors/json_reference/#lifecycle-scripts) - I'll do a second PR to include that once I've got this accepted.

The further thought occurs that maybe a future refactoring could be done to use the [JSON schema](https://github.com/devcontainers/spec/blob/main/docs/specs/devcontainer-reference.md) to ensure validity with the standard. I've got a couple of ideas on how this could be done, but this is probably something that should be specified fully by a maintainer first as it's likely to be a big old change.

mrsimonemms:
Hey @pascalbreuninger, I've had another bash at this based on your feedback. Let me know what you think

FabianKramm:
@mrsimonemms thanks a lot for the progress on this! We support the initialize command, although its at a different place: https://github.com/loft-sh/devpod/blob/87a9b221946941a3a7f27f6ac1e5c334a10711f5/pkg/devcontainer/run.go#L170 .

FabianKramm (approved):
@mrsimonemms thanks a lot, LGTM!

pascalbreuninger:
@mrsimonemms thanks for the contribution and patience! LGTM as well

mrsimonemms:
@pascalbreuninger I'd gone far enough into this one that I was determined to fix it 🤣

PR #499, opened 2023-07-11
fix: do not error out too soon when searching for docker-compose

### LINKED ISSUE #379: Fix docker-compose/docker compose check in Path
Currently we check for both in users' Path but the check fails after docker-compose check.

We should continue and fail if only none is found.

89luca89:
Fix #379 
Resolves ENG-1526

This solves the issue where the docker compose helper only searches for `docker-compose` while ignoring the (working) `docker compose` command

### DIFF
--- a/pkg/compose/helper.go
+++ b/pkg/compose/helper.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


FabianKramm (approved):

FabianKramm:
@89luca89 thanks a lot! LGTM!

PR #520, opened 2023-07-14
fix: escape function after successful clone

p5:
Introduced in https://github.com/loft-sh/devpod/commit/350fac9f6ff85a825559ecbe809ef3e3ae8245f6 was a logic error where if `CloneRepository` succeeds without an error, the function is not exited and is caught by the catch-all 
```
return fmt.Errorf("either workspace repository, image or local-folder is required")
```

This PR updates the function so that if there's no error returned from `CloneRepository`, the function exits.


p5:
Sorry, I'm terrible at naming stuff.  If you know of a better title, feel free to change it :)

pascalbreuninger:
@p5 thanks for the PR, makes sense

PR #551, opened 2023-07-21
Ensure correct uid and gid for remoteUser based on the workspace owner

### LINKED ISSUE #498: DevPod doesn't support local features
I have

```
  "features": {
 ...
    "./local-subdir1": {},
    "./local-subdir2": {}
  }
```

`devcontainer build` supports this syntax. Devpod doesn't, it just prints `20:08:56 info Creating devcontainer...` and exits immediately

### LINKED ISSUE #601: UID in container is not updated to match the host
**What happened?**  
While testing DevPod I've noticed that the UID of the `remoteUser` does not match the UID of the host user. This causes the classic file permission issues between the host and the container.

The UID of the container to match the UID of the host user.

My UID on my host is very large due to logins being controlled by our IT department / active directory.
`uid=1788609210`
`gid=1788600513`
I suspect this may have something to do with it.
I'm using the default out of the box `vscode-remote-try-cpp` sample that I setup through the DevPod GUI using the create workspace "Or select one of our quickstart example". When I ssh in or connect via VS Code I run `id` and see `uid=1000(vscode) gid=1000(vscode) groups=1000(vscode),999(vcpkg)`. When I `cd` to it on my host and run `~/.devpod/agent/contexts/default/workspaces/vscode-remote-try-cpp$ ls -al`
I get:
```
drwxr-xr-x 3 ricky domain^users 4096 Aug  9 22:10 .
drwxr-xr-x 3 ricky domain^users 4096 Aug  9 22:10 ..
drwxr-xr-x 5              1000         1000 4096 Aug  9 22:10 content
-rw-r--r-- 1 ricky domain^users 1155 Aug  9 22:14 workspace.json
```
You can see the issue is the content workspace is owned by a non-existing user on my host giving me file permission issues. I've played around with this outside of the sample setting the `remoteUser` and `updateRemoteUserUID` to true (which it should already default to according to the documentation  https://containers.dev/implementors/json_reference/) but have had no luck.

**Local Environment:**  
- DevPod Version: v0.3.4
- Operating System: Ubuntu 22.04
- ARCH of the OS: AMD64

**DevPod Provider:**  

neogopher:
With this PR, we update the UID of the remote user to match that of the host user. This is done to overcome the classic file permissions issue.
This change is applied only if the host OS is linux or the driver is kubernetes and the host UID is different than the remote user's UID.

P.S. In case of certain IDE's like openvscode, you may need to recreate the devcontainer for the changes to take effect. This can be done using the below command.
```
devpod up <URL-or-local-path> --recreate
```

Fixes #498 
Fixes ENG-1720
Fixes #601 
Fixes ENG-1854

### DIFF
--- a/e2e/tests/up/up.go
+++ b/e2e/tests/up/up.go
@@ -348,6 +348,117 @@ var _ = DevPodDescribe("devpod up test suite", func() {
 				err = f.DevPodUp(ctx, tempDir)
 				framework.ExpectNoError(err)
 			}, ginkgo.SpecTimeout(60*time.Second))
+			ginkgo.It("should start a new workspace with non-root workspace owner and default remoteUser", func(ctx context.Context) {
+				tempDir, err := framework.CopyToTempDir("tests/up/testdata/docker")
+				framework.ExpectNoError(err)
+				ginkgo.DeferCleanup(framework.CleanupTempDir, initialDir, tempDir)
+
+				hostUID, hostGID := 1001, 1001
+
+				err = os.Lchown(tempDir, hostUID, hostGID)
+				framework.ExpectNoError(err)
+
+				f := framework.NewDefaultFramework(initialDir + "/bin")
+
+				_ = f.DevPodProviderDelete(ctx, "docker")
+				err = f.DevPodProviderAdd(ctx, "docker")
+				framework.ExpectNoError(err)
+				err = f.DevPodProviderUse(context.Background(), "docker")
+				framework.ExpectNoError(err)
+
+				ginkgo.DeferCleanup(f.DevPodWorkspaceDelete, context.Background(), tempDir)
+
+				// Wait for devpod workspace to come online (deadline: 30s)
+				err = f.DevPodUp(ctx, tempDir)
+				framework.ExpectNoError(err)
+
+				workspace, err := f.FindWorkspace(ctx, tempDir)
+				framework.ExpectNoError(err)
+
+				projectName := workspace.ID
+
+				containerUID, _, err := f.ExecCommandCapture(ctx, []string{"ssh", "--command", "id -u | xargs echo -n", projectName})
+				framework.ExpectNoError(err)
+				gomega.Expect(containerUID).To(gomega.Equal(fmt.Sprint(hostUID)))
+
+				containerGID, _, err := f.ExecCommandCapture(ctx, []string{"ssh", "--command", "id -g | xargs echo -n", projectName})
+				framework.ExpectNoError(err)
+				gomega.Expect(containerGID).To(gomega.Equal(fmt.Sprint(hostGID)))
+			}, ginkgo.SpecTimeout(60*time.Second))
+			ginkgo.It("should start a new workspace with non-root workspace owner and root remoteUser", func(ctx context.Context) {
+				tempDir, err := framework.CopyToTempDir("tests/up/testdata/docker-root")
+				framework.ExpectNoError(err)
+				ginkgo.DeferCleanup(framework.CleanupTempDir, initialDir, tempDir)
+
+				hostUID, hostGID := 1001, 1001
+
+				err = os.Lchown(tempDir, hostUID, hostGID)
+				framework.ExpectNoError(err)
+
+				f := framework.NewDefaultFramework(initialDir + "/bin")
+
+				_ = f.DevPodProviderDelete(ctx, "docker")
+				err = f.DevPodProviderAdd(ctx, "docker")
+				framework.ExpectNoError(err)
+				err = f.DevPodProviderUse(context.Background(), "docker")
+				framework.ExpectNoError(err)
+
+				ginkgo.DeferCleanup(f.DevPodWorkspaceDelete, context.Background(), tempDir)
+
+				// Wait for devpod workspace to come online (deadline: 30s)
+				err = f.DevPodUp(ctx, tempDir)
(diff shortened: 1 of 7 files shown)


89luca89 (approved):

FabianKramm (commented):
@neogopher thanks for the PR! My question is, if we do it like this, why do we have to still chown the workspace in the setup container?

pkg/devcontainer/setup/setup.go
```diff
@@ -144,6 +152,89 @@ func ChownWorkspace(setupInfo *config.Result, log log.Logger) error {
...
+	// get container uid and gid
+	containerUserName := config.GetRemoteUser(setupInfo)
+	containerUser, err := user.Lookup(containerUserName)
+	if err != nil {
+		return errors.Wrap(err, "get container user")
+	}
+	containerUID, containerGID := containerUser.Uid, containerUser.Gid
+
+	// if host or container user is root or host and container uid/gid match already, do nothing
+	if hostUID == "0" || containerUID == "0" || (hostUID == containerUID && hostGID == containerGID) {
```
FabianKramm, line 175: What happens if the host is not root, but within the container the user is root? Aren't we trying to map the root user to a different UID, GID then?

FabianKramm:
I think we need a couple of tests around this as I believe there will be cases where this will cause issues

FabianKramm (changes requested):
Please add more tests for this (especially with container being root and host not root)

pkg/devcontainer/setup/setup.go
89luca89: I think here we should just check if containerUID is 0, as that's the regular way the user is mapped inside the container
Just dropping the hostUID check

89luca89:
> @neogopher thanks for the PR! My question is, if we do it like this, why do we have to still chown the workspace in the setup container?

I think we can drop that, I'm trying to think of a situation where that is needed, but can't come up with one

pkg/devcontainer/setup/setup.go
neogopher: Yes Fabian, when the hostUID is non-root and the containerUID is 0 (root), we don't do anything; else this will overwrite the UID and GID within the container to hostUID/GID making it just a normal user account within the container.

neogopher:
Hi @FabianKramm and @89luca89,
I have dropped the hostUID check, removed the chownWorkspace and added a few e2e tests to cover this.

e2e/tests/up/up.go
```diff
@@ -311,6 +315,80 @@ var _ = DevPodDescribe("devpod up test suite", func() {
 				err = f.DevPodUp(ctx, tempDir)
 				framework.ExpectNoError(err)
 			}, ginkgo.SpecTimeout(60*time.Second))
+			ginkgo.It("should start a new workspace with non-root workspace owner and default remoteUser", func(ctx context.Context) {
```
89luca89, line 318: An additional tests we can add is to have a custom dockerfile with a vscode user with a UID not-1000 (like 1007 I don't know) and then ensure also that this works and that after the "up" we have a proper vscode user with the right UID

89luca89 (changes requested):

89luca89:
Just a small comment to add a test, for the rest LGTM :+1:

e2e/tests/up/up.go
neogopher: Sure. Nice idea!! I will add this test.

89luca89 (approved):

pkg/devcontainer/setup/setup.go
```diff
@@ -22,18 +24,17 @@ const (
...
 
-func SetupContainer(setupInfo *config.Result, workspaceInfo *provider2.AgentWorkspaceInfo, chownWorkspace bool, log log.Logger) error {
-	// write result to ResultLocation
-	WriteResult(setupInfo, log)
-
-	// chown user dir
-	if chownWorkspace {
-		err := ChownWorkspace(setupInfo, log)
+func SetupContainer(setupInfo *config.Result, workspaceInfo *provider2.AgentWorkspaceInfo, log log.Logger) error {
+	if runtime.GOOS == "linux" || workspaceInfo.Agent.Driver == provider2.KubernetesDriver {
```
FabianKramm, line 28: I don't think this is correct now, the problem is that runtime.GOOS will now always be true, since this is executed within the container

FabianKramm (changes requested):

aacebedo:
This PR solves a problem I also have with the chown when using podman. Current version changes the ownership on the workspace and make it unreachable on the host side.

@neogopher or @FabianKramm Is this going to be merged?

neogopher:
Hi @aacebedo :wave: ,

Yes, this PR is still under review and should be available in the upcoming devpod release once it is verified and merged.

tse77:
I'm too experiencing this issue and is somewhat a blocker for us in order to use this tool successfully. Is there is any chance this will be fixed any time soon?

pascalbreuninger:
closing due to inactivity, feel free to reopen

euven:
I'm not sure why an MR that is seems to be close is getting closed? Can we please re-open, as this is addressing a real issue: #601

euven:
@FabianKramm is there a reason this didn't merge? Thanks for a great tool btw :)

StefanNeubert:
+1 having the same issue with active directory users on linux in a domain

PR #569, opened 2023-07-26
Allow checking out from commit hash

neogopher:
Closes ENG-1784

### DIFF
--- a/cmd/helper/get_workspace_config.go
+++ b/cmd/helper/get_workspace_config.go
@@ -146,14 +146,18 @@ func findDevcontainerFiles(ctx context.Context, rawSource, tmpDirPath string, ma
 	}
 
 	// git repo
-	gitRepository, gitBranch := git.NormalizeRepository(rawSource)
+	gitRepository, gitBranch, gitCommit := git.NormalizeRepository(rawSource)
 	if strings.HasSuffix(rawSource, ".git") || git.PingRepository(gitRepository) {
 		log.Debug("Git repository detected")
 		result.IsGitRepository = true
 
 		log.Debugf("Cloning git repository into %s", tmpDirPath)
 		// git clone --bare --depth=1 $REPO
-		cloneArgs := []string{"clone", "--bare", "--depth=1", gitRepository, tmpDirPath}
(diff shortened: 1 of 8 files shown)


FabianKramm (commented):
@neogopher thanks a lot for the PR! I think you will also need to add it to https://github.com/loft-sh/devpod/pull/569/files#diff-62b624ff068a04565ac9f14b9c4022d9b1a8ea2b02ec6fb660100abc69137202R146 , as that is checking for the GitBranch as well

pkg/workspace/workspace.go
```diff
@@ -443,7 +444,7 @@ func resolve(
 		return workspace, nil
 	}
 
-	return nil, fmt.Errorf("%s is neither a local folder, git repository or docker image", name)
+	return nil, fmt.Errorf("%s is neither a local folder, git repository or docker image\n%#v %#v", name, gitRepository, gitBranch)
```
FabianKramm, line 447: Is this a debug print?

cmd/helper/get_workspace_config.go
```diff
@@ -163,6 +163,19 @@ func findDevcontainerFiles(ctx context.Context, rawSource, tmpDirPath string, ma
 		}
 		log.Debug("Done cloning git repository")
 
+		if gitCommit != "" {
```
FabianKramm, line 166: Is this working? Because when we clone with --depth=1, doesn't that cause any issues?

pkg/workspace/workspace.go
neogopher: Yes. Removed that now.

cmd/helper/get_workspace_config.go
neogopher: No I checked now, and this isn't working. If we wish to reach to the commit, we will have to either pull the complete repo or play around with the depth using 'git deepen' to reach the commit we want. I feel we can remove the resetting to this commit and just check the head of the clone to verify whether the repository is valid. What do you think?

89luca89 (approved):

FabianKramm (approved):

PR #573, opened 2023-07-27
Fix pod manifest template

titilambert:
Hello,

I just tested the latest release, and the pod manifest template feature was not working as expected.
the env vars and volume mounts were not kept from the template.

This patch fix it


89luca89 (approved):

FabianKramm:
@titilambert thanks for the PR! You are correct, we refactored it a bit, but looks like there was a mistake, thanks!

PR #585, opened 2023-08-02
feat: add devcontainer-image flag to override container image in workspaces

### LINKED ISSUE #299: feature: custom image for docker provider
Add a configuration in the docker provider to use a custom image instead of the one that is inferred or choosen by devcontainer.json

This could be useful in case the user uses personalized development images that wants to keep consistent between multiple projects without necessarly modifying the devcontainer.json in a repo

89luca89:
This PR adds support for a `--devcontainer-image` flag in the `up` command that will let the user override the container image with whatever image they want

Fix #299 
Resolves ENG-1417


e2e/tests/up/up.go
```diff
@@ -503,6 +503,34 @@ var _ = DevPodDescribe("devpod up test suite", func() {
...
+				tempDir, err := framework.CopyToTempDir("tests/up/testdata/docker")
+				framework.ExpectNoError(err)
+				ginkgo.DeferCleanup(framework.CleanupTempDir, initialDir, tempDir)
+
+				f := framework.NewDefaultFramework(initialDir + "/bin")
+
+				_ = f.DevPodProviderDelete(ctx, "docker")
+				err = f.DevPodProviderAdd(ctx, "docker")
+				framework.ExpectNoError(err)
+				err = f.DevPodProviderUse(context.Background(), "docker")
```
ThomasK33, line 516: Shouldn't this also use `ctx`?

```suggestion
				err = f.DevPodProviderUse(ctx, "docker")
```

ThomasK33 (commented):
LGTM

ThomasK33 (approved):

pbialon (approved):

pkg/devcontainer/run.go
```diff
@@ -106,6 +119,11 @@ func (r *Runner) prepare() (*config.SubstitutedConfig, *WorkspaceConfig, error)
 	if parsedConfig.WorkspaceMount != "" {
 		workspace.WorkspaceMount = parsedConfig.WorkspaceMount
 	}
+
+	if options.DevContainerImage != "" {
+		parsedConfig.ImageContainer = config.ImageContainer{Image: options.DevContainerImage}
```
FabianKramm, line 124: What happens when in the devcontainer.json a Dockerfile build is specified? Does this still work then?

89luca89: It was ignored, I've fixed it now :+1:

FabianKramm (approved):

PR #588, opened 2023-08-02
Rebuild on changes to files in build context

neogopher:
With this PR, we will build a hash for the contents in the build context and append it to the Prebuild Hash. This will cause any changes to files under the build context to be picked up while rebuilding the workspace.

Fixes ENG-1797


FabianKramm (approved):
@neogopher thanks a lot for the PR! LGTM!

Issue #507, opened 2023-07-12
Using env variables in the devcontainer.json


misialq:
**What happened?**
I keep getting an error when trying to create a workspace using the SSH provider. Here is the relevant section of the log:

```
[11:24:40] info docker: invalid reference format.
[11:24:40] info exit status 125
[11:24:40] info start dev container
[11:24:40] info github.com/loft-sh/devpod/pkg/devcontainer.(*Runner).runSingleContainer
[11:24:40] info /home/runner/work/devpod/devpod/pkg/devcontainer/single.go:75
[11:24:40] info github.com/loft-sh/devpod/pkg/devcontainer.(*Runner).Up
[11:24:40] info /home/runner/work/devpod/devpod/pkg/devcontainer/run.go:130
[11:24:40] info github.com/loft-sh/devpod/cmd/agent/workspace.(*UpCmd).devPodUp
[11:24:40] info /home/runner/work/devpod/devpod/cmd/agent/workspace/up.go:373
[11:24:40] info github.com/loft-sh/devpod/cmd/agent/workspace.(*UpCmd).up
[11:24:40] info /home/runner/work/devpod/devpod/cmd/agent/workspace/up.go:167
[11:24:40] info github.com/loft-sh/devpod/cmd/agent/workspace.(*UpCmd).Run
[11:24:40] info /home/runner/work/devpod/devpod/cmd/agent/workspace/up.go:101
[11:24:40] info github.com/loft-sh/devpod/cmd/agent/workspace.NewUpCmd.func1
[11:24:40] info /home/runner/work/devpod/devpod/cmd/agent/workspace/up.go:61
[11:24:40] info github.com/spf13/cobra.(*Command).execute
[11:24:40] info /home/runner/work/devpod/devpod/vendor/github.com/spf13/cobra/command.go:916
[11:24:40] info github.com/spf13/cobra.(*Command).ExecuteC
[11:24:40] info /home/runner/work/devpod/devpod/vendor/github.com/spf13/cobra/command.go:1044
[11:24:40] info github.com/spf13/cobra.(*Command).Execute
[11:24:40] info /home/runner/work/devpod/devpod/vendor/github.com/spf13/cobra/command.go:968
[11:24:40] info github.com/loft-sh/devpod/cmd.Execute
[11:24:40] info /home/runner/work/devpod/devpod/cmd/root.go:89
[11:24:40] info main.main
[11:24:40] info /home/runner/work/devpod/devpod/main.go:8
[...]

PR #591, opened 2023-08-03
Prepared documentation for using env variables in devcontainer json

pbialon:
Fixes ENG-1729

A user encountered some problems with a setup of his workspace when using env variables in devcontainer.json and passing them through the ssh: https://github.com/loft-sh/devpod/issues/507 

I'm adding here an instruction which I send him to verify he didn't forgot about anything.

### DIFF
--- a/docs/pages/developing-in-workspaces/environment-variables-in-devcontainer-json.mdx
+++ b/docs/pages/developing-in-workspaces/environment-variables-in-devcontainer-json.mdx
@@ -1,0 +1,43 @@
+---
+title: Environment variables in devcontainer.json
+sidebar_label: Environment variables in devcontainer.json
+---
+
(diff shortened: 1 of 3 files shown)


pascalbreuninger (approved):
LGTM! One additional section you could add while we're at it is how ppl can use `--workspace-env` and `${containerEnv:FOO}` in their `devcontainer.json`. Not too important though

pbialon:
> LGTM! One additional section you could add while we're at it is how ppl can use `--workspace-env` and `${containerEnv:FOO}` in their `devcontainer.json`. Not too important though

I spoke with @FabianKramm and I'm not sure I fully understand the behaviour of this flag. From one side it looks like the env variables passed through `--workspace-env` are accessible with `${containerEnv:}` directive, but they are accessible on the remoteHost (so outside the container). I don't want to mislead anybody so I won't include the instruction for that.

PR #592, opened 2023-08-06
fix: cannot launch vscode

### LINKED ISSUE #460: Can't connect to devpod using VS Code
**What happened?**

devpod can't connect to vs code

use vscode with devpod

I just created the python container and got the error after it launched VS Code

**Local Environment:**

* DevPod Version: v0.1.10
* Operating System: mac
* ARCH of the OS: ARM64

**DevPod Provider:**

* Local/remote provider: docker

image

cakriwut:
Fix #460 - cannot launch vscode.


FabianKramm:
@cakriwut thanks a lot for the PR! I think I accidentally fixed this already with #594, would you mind checking if this also fixes your problem? Its available in v0.3.4

cakriwut:
It works, thanks.
However I found another bug that happen using remote Kubernetes.

PR #598, opened 2023-08-08
ENG-1823 fix path env variable

pbialon:
If the PATH variable contained new lines we were parsing it improperly, because of the bug in the upstream library (issue related: https://github.com/tauri-apps/fix-path-env-rs/issues/7). 

Here, I reused most of the `fix-path-env-rs` code. But instead using `env` command and then parse all the variables and get value for `PATH` only, I use `printenv PATH` command to get only `PATH` value and simplify processing.

Apart from that I decided to split the code into multiple functions for more readability and added two simple testcases. 

I tested it locally and managed to create kubernetes provider in Devpod Desktop.

fixes ENG-1823 
fixes ENG-1824


pascalbreuninger (approved):
LGTM! thank you

PR #599, opened 2023-08-09
Feat/gpg agent forwarding

### LINKED ISSUE #580: Request: sync gpg signing keys in addition to git & docker credentials
For projects that utilize commit signing ([https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work](https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work)), it would be really handy to have GPG signing keys already available in the workspace.

No

Coder appears to have this option ([https://coder.com/docs/v2/latest/cli/ssh#-g---forward-gpg](https://coder.com/docs/v2/latest/cli/ssh#-g---forward-gpg)) and appears to implement it by first uploading GPG keys and then forwarding the GPG Agent: [https://github.com/coder/coder/blob/af036b440bd776f5fe57e0ddea7fc0c0596ffff7/cli/ssh.go#L297C8-L297C8](https://github.com/coder/coder/blob/af036b440bd776f5fe57e0ddea7fc0c0596ffff7/cli/ssh.go#L297C8-L297C8).

It looks like it may be possible to forward the GPG Agent via SSH config:
[https://unix.stackexchange.com/questions/656738/how-to-forward-gpg-agent-from-host-to-docker-container-and-reuse-cached-passphra](https://unix.stackexchange.com/questions/656738/how-to-forward-gpg-agent-from-host-to-docker-container-and-reuse-cached-passphra).

**Additional context**

89luca89:
This PR Will

- migrate to our ssh server library github.com/loft-sh/ssh
- add reverse-forward-ports to ssh command
- add gpg-agent-forwarding to up command

Before merging I'd like to have feedbacks from non-linux users, as I've been able to test this only on a series of linux machines

Fixes #580 
Resolves ENG-1809


cmd/up.go
```diff
@@ -228,6 +240,15 @@ func (cmd *UpCmd) Run(
 		}
 	}
 
+	// if GpgAgentForwarding we need to keep running in order to keep the reverse tunnel
+	// functioning, or gpg-agent will drop
+	if cmd.GpgAgentForwarding {
+		log.Infof(
+			"GPG Agent forwarding specified, keep this process running to have working gpg-agent forwarding",
+		)
+		select {}
```
FabianKramm, line 249: Why not call this at the end of up and then block?

FabianKramm: Would also be good to have a context variable for this so we don't need to use the flag everytime

cmd/up.go
```diff
@@ -819,3 +840,222 @@ func setupDotfiles(
...
+echo $agent_socket
+if [ -S $agent_socket ]; then
+  echo agent socket exists, attempting to kill it >&2
+  gpgconf --kill gpg-agent
+  rm -f $agent_socket
+  sleep 1
+fi
+
+test ! -S $agent_socket
+'`)
```
FabianKramm, line 900: Instead of having this bash script, can we create a devpod command for this, e.g. `devpod agent workspace start-gpg-agent`

cmd/up.go
```diff
@@ -819,3 +840,222 @@ func setupDotfiles(
...
+
+	err = exec.Command(execPath, fixAgent...).Run()
+	if err != nil {
+		return err
+	}
+
+	log.Infof("gpg-agent forwarding done")
+
+	return nil
```
FabianKramm, line 1060: can we condense all this into a single command that is executed? so the goal should be to only call exec.Command once

pkg/ssh/forward.go
```diff
@@ -30,7 +36,9 @@ func PortForward(ctx context.Context, client *ssh.Client, localNetwork, localAdd
 	}()
 
 	counter := newConnectionCounter(ctx, exitAfterTimeout, func() {
-		log.Fatalf("Stopping devpod up, because it stayed idle for a while. You can disable this via 'devpod context set-options -o EXIT_AFTER_TIMEOUT=false'")
+		log.Fatalf(
+			"Stopping devpod up, because it stayed idle for a while. You can disable this via 'devpod context set-options -o EXIT_AFTER_TIMEOUT=false'",
```
FabianKramm, line 40: Can you revert that?

pkg/ssh/forward.go
```diff
@@ -86,3 +99,87 @@ func forward(localConn net.Conn, client *ssh.Client, remoteNetwork, remoteAddr s
...
+		select {
+		case <-done:
+		case <-ctx.Done():
+			_ = listener.Close()
+		}
+	}()
+
+	counter := newConnectionCounter(ctx, exitAfterTimeout, func() {
+		log.Fatalf(
+			"Stopping devpod up, because it stayed idle for a while. You can disable this via 'devpod context set-options -o EXIT_AFTER_TIMEOUT=false'",
```
FabianKramm, line 129: That looks very weird, can you do a log.Fatal instead

FabianKramm (changes requested):

pkg/ssh/forward.go
89luca89: Done

pkg/ssh/forward.go
89luca89: Done

cmd/up.go
89luca89: Done

cmd/up.go
89luca89: Done

89luca89:
Tested working on Macos also :+1:

docs/pages/developing-in-workspaces/credentials.mdx
```diff
@@ -23,4 +23,20 @@ DevPod will make docker registry credentials available inside the dev container
 If you don't want DevPod to inject the credentials, you can disable that via the following command for all workspaces:
 ```
 devpod context set-options default -o INJECT_DOCKER_CREDENTIALS=false
-```
\ No newline at end of file
+```
+
+## GPG credentials
+
+DevPod will gpg keys available inside the dev container through an ssh tunnel. This allows you to sign commits from inside the workspace.
```
pbialon, line 30: "will gpg keys"? will forward?

pkg/ssh/forward.go
```diff
@@ -86,3 +97,85 @@ func forward(localConn net.Conn, client *ssh.Client, remoteNetwork, remoteAddr s
...
+	defer listener.Close()
+
+	done := make(chan struct{})
+	defer close(done)
+
+	go func() {
+		select {
+		case <-done:
+		case <-ctx.Done():
+			_ = listener.Close()
```
pbialon, line 121: this function looks exactly the same as `PortForward()` - only differs in calling `forward()` and `reverseForward()`. 

What about not duplicating the code, by reuse it?

pbialon: And those two (forward/reverseForward()) also look similar. Maybe we don't need to duplicate that code as well.

cmd/ssh.go
```diff
@@ -192,16 +226,78 @@ func (cmd *SSHCmd) jumpContainer(ctx context.Context, devPodConfig *config.Confi
...
+				ctx,
+				devPodConfig,
+				containerClient,
+				client.WorkspaceConfig().IDE.Name,
+				log,
+			)
+		})
+}
+
+func (cmd *SSHCmd) reverseForwardPorts(
```
pbialon, line 245: `reverseForwardPorts()` also looks really similar to `forwardPorts`

cmd/ssh.go
```diff
@@ -233,10 +344,31 @@ func (cmd *SSHCmd) forwardPorts(ctx context.Context, containerClient *ssh.Client
 	return <-errChan
 }
 
-func (cmd *SSHCmd) startTunnel(ctx context.Context, devPodConfig *config.Config, containerClient *ssh.Client, ideName string, log log.Logger) error {
+func (cmd *SSHCmd) startTunnel(
```
pbialon, line 347: in general those kind of refactorings I'd put in separate PR - there is already a lot of lines of code :)

cmd/ssh.go
```diff
@@ -284,3 +456,89 @@ func (cmd *SSHCmd) startServices(ctx context.Context, devPodConfig *config.Confi
 		}
 	}
 }
+
+// setupGPGAgent will forward a local gpg-agent into the remote container
+// this works by using cmd/agent/workspace/setup_gpg
+func (cmd *SSHCmd) setupGPGAgent(
+	ctx context.Context,
+	devPodConfig *config.Config,
```
pbialon, line 464: unused devPodConfig

cmd/ssh.go
```diff
@@ -284,3 +456,89 @@ func (cmd *SSHCmd) startServices(ctx context.Context, devPodConfig *config.Confi
 		}
 	}
 }
+
+// setupGPGAgent will forward a local gpg-agent into the remote container
+// this works by using cmd/agent/workspace/setup_gpg
+func (cmd *SSHCmd) setupGPGAgent(
```
pbialon, line 462: if you have an idea, you can split that function - it's a bit longish :)

pbialon (dismissed):

docs/pages/developing-in-workspaces/credentials.mdx
89luca89: Good catch :)

pkg/ssh/forward.go
89luca89: Yep done now for Forward/ReverseForward

But the actual forwarding functions (forward/reverse) have quite separate concepts and connections, so I'd like to keep them separated for the sake of simplicity

cmd/ssh.go
89luca89: As before yes they are similar but separate (opposite) concepts and I'd like to keep them separated

cmd/ssh.go
89luca89: Thanks done :+1:

cmd/ssh.go
89luca89: Yes it's a bit on the long side, but I don't feel like it's doing out-of-scope things, it's doing that one single thing
which happens to be a bit long :joy:

cmd/ssh.go
```diff
@@ -248,8 +380,34 @@ func (cmd *SSHCmd) startTunnel(ctx context.Context, devPodConfig *config.Config,
...
+
+		if err == nil && len(out.Bytes()) > 1 {
+			log.Debugf("gpg: exporting already running, skipping")
+		} else {
+			err := cmd.setupGPGAgent(ctx, devPodConfig, containerClient, log)
+			if err != nil {
+				return err
+			}
+		}
+	}
```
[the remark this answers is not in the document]
89luca89: That would be a separate function called only by this function, I don't feel like it makes a lot of sense here

cmd/agent/workspace/setup_gpg.go
```diff
@@ -0,0 +1,201 @@
...
+
+// will forward a local gpg-agent into the remote container
+// this works by
+//
+// - stopping remote gpg-agent and removing the sockets
+// - exporting local public keys and owner trust
+// - importing those into the container
+// - ensuring the gpg-agent is stopped in the container
+// - starting a reverse-tunnel of the local unix socket to remote
+// - ensuring paths and permissions are correctly set in the remote
```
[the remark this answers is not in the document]
89luca89: Nice idea, done :smile:

cmd/agent/workspace/setup_gpg.go
```diff
@@ -0,0 +1,230 @@
...
+		return err
+	}
+
+	time.Sleep(time.Second)
+
+	return nil
+}
+
+func (cmd *SetupGPGCmd) stopGpgAgent() error {
+	return exec.Command("gpgconf", []string{"--kill", "gpg-agent"}...).Run()
```
pascalbreuninger, line 122: does this fail if the agent is already stopped?

cmd/ssh.go
```diff
@@ -284,3 +456,88 @@ func (cmd *SSHCmd) startServices(ctx context.Context, devPodConfig *config.Confi
...
+		gpgExtraSocketPath,
+	)
+
+	command := strings.Join(forwardAgent, " ")
+
+	if cmd.User != "" && cmd.User != "root" {
+		command = fmt.Sprintf("su -c \"%s\" '%s'", command, cmd.User)
+	}
+
+	// We use start to keep this connection alive in background (hence the use of sleep infinity)
```
pascalbreuninger, line 541: This comment seems to be somewhat detached from the code, could you either elaborate or rewrite pls?

cmd/agent/workspace/setup_gpg.go
```diff
@@ -0,0 +1,230 @@
...
+// will forward a local gpg-agent into the remote container
+// this works by
+//
+// - stopping remote gpg-agent and removing the sockets
+// - exporting local public keys and owner trust
+// - importing those into the container
+// - ensuring the gpg-agent is stopped in the container
+// - starting a reverse-tunnel of the local unix socket to remote
+// - ensuring paths and permissions are correctly set in the remote
+func (cmd *SetupGPGCmd) Run(ctx context.Context) error {
```
pascalbreuninger, line 55: nit: feels like there's enough in here to justify a `pkg/gpg`

cmd/ssh.go
pascalbreuninger, line 245: this could potentially be invoked twice with similar arguments, is it safe to do so?

cmd/agent/workspace/setup_gpg.go
89luca89: Nope and that's ok we want to have it stopped anyway

cmd/ssh.go
89luca89: Ah it's a leftover from a previous implementation, I'll remove it

cmd/ssh.go
89luca89: What do you mean? as in "run multiple times the gpg-forwarding"?

cmd/agent/workspace/setup_gpg.go
89luca89: Done

cmd/ssh.go
89luca89: Removed

cmd/ssh.go
pascalbreuninger: nope, but if you provide the flag **and** gpg forwarding, this function will be called twice (or I missed smth)

pkg/ssh/server/ssh.go
```diff
@@ -50,13 +51,28 @@ func NewServer(addr string, hostKey []byte, keys []ssh.PublicKey, log log.Logger
...
 			},
+			ReverseUnixForwardingCallback: func(ctx ssh.Context, socketPath string) bool {
+				log.Debugf("attempt to bind socket %s", socketPath)
+
+				_, err := os.Stat(socketPath)
+				if err == nil {
+					log.Debugf("%s already exists, removing", socketPath)
+
+					_ = os.Remove(socketPath)
+				}
```
pbialon, line 62: why do we delete it? And where the binding happens?

pkg/ssh/forward.go
```diff
@@ -86,3 +141,44 @@ func forward(localConn net.Conn, client *ssh.Client, remoteNetwork, remoteAddr s
...
+		}
+	}()
+	waitGroup.Add(1)
+
+	// Copy sshConn.Reader to localConn.Writer
+	go func() {
+		defer waitGroup.Done()
+		defer remoteConn.Close()
+
+		_, err = io.Copy(remoteConn, localConn)
```
pbialon, line 177: in similar cases in `loft-sh/ssh` we were using `bicopy` - why this one is different?

pkg/gpg/gpg_forwarding.go
```diff
@@ -0,0 +1,126 @@
...
+func (g *GPGConf) SetupGpgConf() error {
+	_, err := os.Stat(filepath.Join(os.Getenv("HOME"), ".gnupg", "gpg.conf"))
+	if err != nil {
+		_, err = os.Create(filepath.Join(os.Getenv("HOME"), ".gnupg", "gpg.conf"))
+		if err != nil {
+			return err
+		}
+	}
+
+	gpgConfig, err := os.ReadFile(filepath.Join(os.Getenv("HOME"), ".gnupg", "gpg.conf"))
```
pbialon, line 62: maybe we can have a method `configPath()` returning `filepath.Join(os.Getenv("HOME"), ".gnupg", "gpg.conf")`

pkg/gpg/gpg_forwarding.go
```diff
@@ -0,0 +1,126 @@
...
+
+		if _, err := f.WriteString("use-agent\n"); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (g *GPGConf) SetupRemoteSocketDirtree() error {
```
pbialon, line 83:
```suggestion
func (g *GPGConf) SetupRemoteSocketDirTree() error {
```

pkg/gpg/gpg_forwarding.go
```diff
@@ -0,0 +1,126 @@
...
+		"chown",
+		"-R",
+		strconv.Itoa(os.Getuid())+":"+strconv.Itoa(os.Getgid()),
+		"/run/user",
+		filepath.Dir(g.SocketPath),
+		g.SocketPath,
+	).Run()
+}
+
+func (g *GPGConf) SetupRemoteSocketLink() error {
```
pbialon, line 99: maybe a better structure or comment what that function is doing? It's not clear at the first glance, what is happening to those symlinks etc :)

cmd/agent/workspace/setup_gpg.go
```diff
@@ -0,0 +1,121 @@
...
+		return err
+	}
+
+	logger.Debugf("Setup gpg.conf")
+	err = gpgConf.SetupGpgConf()
+	if err != nil {
+		return err
+	}
+
+	time.Sleep(time.Second)
```
pbialon, line 118: why do we need that sleep?

cmd/ssh.go
```diff
@@ -192,16 +226,78 @@ func (cmd *SSHCmd) jumpContainer(ctx context.Context, devPodConfig *config.Confi
...
+		timeout, err = time.ParseDuration(cmd.ForwardPortsTimeout)
+		if err != nil {
+			return fmt.Errorf("parse forward ports timeout: %w", err)
+		}
 
-		// start ssh tunnel
-		return cmd.startTunnel(ctx, devPodConfig, containerClient, client.WorkspaceConfig().IDE.Name, log)
-	})
+		log.Infof("Using port forwarding timeout of %s", cmd.ForwardPortsTimeout)
+	}
```
pbialon, line 259: move parsing of timeout out to separate function

pkg/ssh/forward.go
```diff
@@ -11,13 +11,63 @@ import (
...
+		remoteNetwork, remoteAddr, localNetwork, localAddr,
+		exitAfterTimeout, log, reverseForward,
+	)
+}
+
+func portForwarding(
+	ctx context.Context,
+	client *ssh.Client,
+	listener net.Listener,
+	srcNetwork, srcAddr, dstNetwork, dstAddr string,
```
pbialon, line 66: `srcNetwork` is not used

cmd/ssh.go
```diff
@@ -192,16 +226,78 @@ func (cmd *SSHCmd) jumpContainer(ctx context.Context, devPodConfig *config.Confi
...
+
+	errChan := make(chan error, len(cmd.ReverseForwardPorts))
```
pbialon, line 261: question: why do we want to have more than one reverse forward port?

Is it because of multiple workspaces using that feature or sth which I don't see?

cmd/ssh.go
```diff
@@ -284,3 +456,87 @@ func (cmd *SSHCmd) startServices(ctx context.Context, devPodConfig *config.Confi
...
+	log log.Logger,
+) error {
+	writer := log.ErrorStreamOnly().Writer(logrus.InfoLevel, false)
+	defer writer.Close()
+
+	log.Debugf("gpg: exporting gpg public key from host")
+
+	// Read the user's public keys and ownertrust from GPG.
+	// These commands are executed LOCALLY, the output will be imported by the remote gpg
+	pubKeyExport, err := exec.Command("gpg", "--armor", "--export").Output()
```
pbialon, line 474: why won't we move those commands (this, and ones below) to gpg package? 

I feel more comfortable with a "client" calling `exec.Command` underneath instead of calling in directly in hear.

cmd/ssh.go
```diff
@@ -248,8 +380,34 @@ func (cmd *SSHCmd) startTunnel(ctx context.Context, devPodConfig *config.Config,
...
+		// Check if a forwarding is already enabled and running, in that case
+		// we skip the forwarding and keep using the original one
+		command := "gpg -K"
+		if cmd.User != "" && cmd.User != "root" {
+			command = fmt.Sprintf("su -c \"%s\" '%s'", command, cmd.User)
+		}
+
+		// capture the output, if it's empty it means we don't have gpg-forwarding
+		var out bytes.Buffer
+		err := devssh.Run(ctx, containerClient, command, nil, &out, writer)
```
pbialon, line 394: maybe encapsulate it somehow, with `remoteGpgClient`? For me it's too low-level abstraction we use for the function which is responsible for checking if gpg forwarding is already set up and if not then to do it.

cmd/ssh.go
89luca89: Tested and it indeed created problems (sometimes)
Fixed it now thanks!

pkg/ssh/server/ssh.go
89luca89: That's the behaviour of OpenSSH with the `StreamLocalBindUnlink=yes`  option
Which is useful in our case to ensure we don't have dangling (old) sockets

pkg/ssh/forward.go
89luca89: Not sure here, might want to ask @FabianKramm (the reverse code is a copy of the regular forward one more or less)

pkg/gpg/gpg_forwarding.go
89luca89: Done :+1:

cmd/agent/workspace/setup_gpg.go
89luca89: Debugging leftover, removed :+1:

pkg/gpg/gpg_forwarding.go
89luca89: Done :+1:

pkg/gpg/gpg_forwarding.go
89luca89: Done :+1:

pkg/ssh/forward.go
89luca89: That's needed in order for `ForwardingFunction` alias/type to work and return the function (eithe regular or reverse forward) to be returned to `portForwarding`

cmd/ssh.go
89luca89: Yes, it may be that one wants to forward more ports, sockets or whatnot at once

cmd/ssh.go
89luca89: Done :+1:

cmd/ssh.go
89luca89: Done :+1:

cmd/ssh.go
89luca89: Moved to the gpg package :+1:

FabianKramm (changes requested):
@89luca89 can we add a unit or e2e test for this? Since this is something thats default disabled and we have changes on the roadmap for the underlying ssh connection, we need to make sure those future changes are not breaking this

89luca89:
> @89luca89 can we add a unit or e2e test for this? Since this is something thats default disabled and we have changes on the roadmap for the underlying ssh connection, we need to make sure those future changes are not breaking this

:+1: sure thing

FabianKramm:
Besides that it looks good, but we need a rebase here

FabianKramm (approved):

Issue #390, opened 2023-06-16
Ports support for kubernetes driver


snapiz:
**Is your feature request related to a problem?**
I'm trying to make workspaces communicate using traefik but can't find a way to expose pod port

**Which solution do you suggest?**
allow to configure pod ports maybe through `portsAttributes` of devcontainers.json ?

**Which alternative solutions exist?**
not able to find a solution

**Additional context**

pascalbreuninger:
Hi @snapiz, thank you for submitting a feature request, we'll look into it 👍

PR #611, opened 2023-08-14
fix: Support Ports and EnvFrom in kubernetes manifest

Joibel:
PR #547 added support for a `podManifestTemplate` which claimed to fix Issue #390 (although not via `portsAttributes` but via the template).

I suspect `ports` in the manifest worked in that PR and were broken by https://github.com/loft-sh/devpod/commit/4be34b237a9653630ae687453b72117908bdc854 which did some good refactoring, but made it necessary to copy the parameters you wanted rather than basing the pod on the manifest directly.

I would like both `ports:` and `envFrom:` to work, and this commit fixes both of those to come from the template. If they are unspecified in the manifest they will be unset in the created pod too.

### DIFF
--- a/pkg/driver/kubernetes/run.go
+++ b/pkg/driver/kubernetes/run.go
@@ -297,6 +297,8 @@ func getContainers(pod *corev1.Pod, imageName, entrypoint string, args []string,
 	}
 	if existingDevPodContainer != nil {
 		devPodContainer.Env = append(existingDevPodContainer.Env, devPodContainer.Env...)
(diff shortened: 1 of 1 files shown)


pascalbreuninger (approved):
Hey @Joibel , thanks for opening the PR! LGTM

Joibel:
> Hey @Joibel , thanks for opening the PR! LGTM

Thanks @pascalbreuninger. The e2e test failures don't look like things I could have broken - are the tests flaky?

pascalbreuninger:
Apparently a bit, re-running worked though 😄

PR #624, opened 2023-08-18
Image from docker compose file

### LINKED ISSUE #606: devpod-cli build does not take image from compose file
**What happened?**  

I have a DevContainer setup using docker compose.
The docker compose file also includes build section to build the image.

Using the following build command, I get an error:

```shell
devpod-cli build $(pwd) --repository my_dummy_repo --skip-push
```

```
15:20:21 info Building devcontainer...
15:20:21 error Error building image: build image: get image build info: get image config remotely: parse reference : could not parse reference:
15:20:21 info build: build image: get image build info: get image config remotely: parse reference : could not parse reference: 
```

If I run the exact same setup directly in the DevPod UI, it spins up the container + IDE.

**Manual workaround**

The workaround is to first parse the image from the compose file, and pass it to build:

```shell
dev_service=$(sed '/^[[:blank:]]*#/d;s/\/\/.*//' .devcontainer/devcontainer.json | jq -r ".service")
compose_image=$(cat .devcontainer/docker-compose.yaml | yq .services.$dev_service.image)
devpod-cli build $(pwd) --repository my_dummy_repo --skip-push --devcontainer-image $compose_image 
```

`devpod-cli build` should take the image from the docker compose file of the referenced service.

My *`devcontainer.json`*:
```json
{
    "name": "Test Compose Build",
    "dockerComposeFile": [
        "docker-compose.yaml"
    ],
    "service": "hello-world-service",
    "customizations": {
        "devpod": {
            "prebuildRepository": "my_hello_world_dev"
        }
    }
}
```

My *`docker-compose.yaml`*:

```yaml
services:
  hello-world-service:
    image: my_hello_world:latest
    build:
      context: ../
      dockerfile: .devcontainer/Dockerfile
    # Overrides default command so things don't shut down after the process ends.
    command: /bin/sh -c "while sleep 1000; do :; done"
```

My *`Dockerfile`*:

```Dockerfile
FROM ubuntu:22.04

RUN apt update && apt install -y curl
RUN echo "Hello World"
```

**Local Environment:**  
- DevPod Version: 0.3.4
- Operating System: linux 
- ARCH of the OS: AMD64

**DevPod Provider:**  

No

neogopher:
With this PR, we enable devpod build to fetch the image name of the referenced service from the service definition in docker-compose file.

P.S. Subsequently pushing the above image is not possible currently and would be discussed and worked upon

Closes ENG-1866
Closes #606

### DIFF
--- a/pkg/devcontainer/compose.go
+++ b/pkg/devcontainer/compose.go
@@ -823,3 +823,7 @@ func mappingToMap(mapping composetypes.MappingWithEquals) map[string]string {
 	}
 	return ret
 }
+
+func isDockerComposeConfig(config *config.DevContainerConfig) bool {
+	return len(config.DockerComposeFile) > 0
+}
--- a/e2e/tests/build/build.go
+++ b/e2e/tests/build/build.go
@@ -98,6 +98,27 @@ var _ = DevPodDescribe("devpod build test suite", func() {
 		framework.ExpectNoError(err)
 	})
 
+	ginkgo.It("should build the image of the referenced service from the docker compose file", func() {
+		ctx := context.Background()
+
+		f := framework.NewDefaultFramework(initialDir + "/bin")
+		tempDir, err := framework.CopyToTempDir("tests/build/testdata/docker-compose")
+		framework.ExpectNoError(err)
+		ginkgo.DeferCleanup(framework.CleanupTempDir, initialDir, tempDir)
+
+		_ = f.DevPodProviderDelete(ctx, "docker")
+		err = f.DevPodProviderAdd(ctx, "docker")
+		framework.ExpectNoError(err)
+		err = f.DevPodProviderUse(context.Background(), "docker")
+		framework.ExpectNoError(err)
+
+		prebuildRepo := "test-repo"
+
+		// do the build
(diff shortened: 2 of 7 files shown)


FabianKramm:
@lizardruss please take a look

pkg/devcontainer/compose.go
```diff
@@ -794,3 +794,7 @@ func mappingToMap(mapping composetypes.MappingWithEquals) map[string]string {
 	}
 	return ret
 }
+
+func isDockerComposeConfig(config *config.DevContainerConfig) bool {
+	return len(config.DockerComposeFile) > 0
+}
```
lizardruss, line 800: Mind also using this function [here](https://github.com/loft-sh/devpod/blob/main/pkg/devcontainer/run.go#L161)?

lizardruss (approved):
LGTM

Only one additional spot where `isDockerComposeConfig()` could be reused, but not mandatory.

pkg/devcontainer/compose.go
neogopher: Thanks @lizardruss . Good catch!! I have updated the code now to reuse the function there.

PR #629, opened 2023-08-21
import workspace cmd arguments

pbialon:
- moved context resolving to separate method
- resolve workspace folder based on other arguments instead of flag
- prepared provider resolver to take care of finding exact provider
- moved Run command up
- cleanup of flags


pascalbreuninger (approved):
LGTM

PR #648, opened 2023-08-27
fix(desktop): add updater back to tauri features

maxbrunet:
I'm trying to package DevPod Desktop with Nix, and building it as regular Rust program with `cargo build` instead of `tauri build` is simpler, but I got this error:

```
The `tauri` dependency features on the `Cargo.toml` file does not match the allowlist defined under `tauri.conf.json`.
Please run `tauri dev` or `tauri build` or add the `updater` feature.
```

It was removed in #616

This Nix PR is here if you'd like to have a look: NixOS/nixpkgs#251858

### DIFF
--- a/desktop/src-tauri/Cargo.toml
+++ b/desktop/src-tauri/Cargo.toml
@@ -15,6 +15,7 @@ serde_json = "1.0"
 serde = { version = "1.0", features = ["derive"] }
 # Tauri
 tauri = { version = "1.2.4", features = [
+        "updater",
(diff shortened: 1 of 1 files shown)


pascalbreuninger:
Hey @maxbrunet, we'd like to keep the `updater` feature out of the `cargo.toml` as it breaks local development if you don't have the correct (sensitive) environment variables. The problem with building it as a standalone rust problem is that you're not going to rebuild the UI

PR #666, opened 2023-08-31
Support checkout from PR Reference

neogopher:
With this PR we support checkout from a PR Reference, so that below should be possible now.

```devpod up <repo-name>@pull/<PR-number>/head```

Closes ENG-1842


89luca89 (approved):
LGTM

FabianKramm (approved):

PR #680, opened 2023-09-04
improvements to error handling in import-workspace

pbialon:
Resolves ENG-1992 

Message for invalid devpod pro host:

Screenshot 2023-09-05 at 13 28 59

Message for not-existing workspace:
Screenshot 2023-09-05 at 13 29 45

Unsupported host:
Screenshot 2023-09-05 at 14 01 46

There is one last case which left - if we lack some id / parameter, we get only "invalid query" error. I spend some time debugging it, but couldn't figure out why the message is not passed to the UI. I'll spend some more time on that, but don't want to block the whole process by only this one case.

### DIFF
--- a/pkg/client/clientimplementation/proxy_client.go
+++ b/pkg/client/clientimplementation/proxy_client.go
@@ -337,7 +337,12 @@ func (s *proxyClient) ImportWorkspace(ctx context.Context, options client.Import
 	s.m.Lock()
 	defer s.m.Unlock()
 
-	stdout := &bytes.Buffer{}
+	reader, writer := io.Pipe()
+	defer writer.Close()
+	go func() {
+		readLogStream(reader, s.log)
+	}()
+
 	buf := &bytes.Buffer{}
 
 	cmd := s.config.Exec.Proxy.ImportWorkspace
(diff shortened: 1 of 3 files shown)


pkg/client/clientimplementation/proxy_client.go
```diff
@@ -337,34 +337,47 @@ func (s *proxyClient) ImportWorkspace(ctx context.Context, options client.Import
 	s.m.Lock()
 	defer s.m.Unlock()
 
-	reader, writer := io.Pipe()
-	defer writer.Close()
-	go func() {
-		readLogStream(reader, s.log)
```
89luca89, line 343: I think this was needed to have the logs from the command forwarded to the current logging level
So that using the `cli` would have the logs follow along

From what I see this is not going to be the case anymore, as now we only print an error at the end (if any is present) but we lose the rest

Maybe we should find a way to have both to `readLogStream` and to a buffer too, so preserving the old functionality but having also the new (correct) behaviour

89luca89 (changes requested):

pkg/client/clientimplementation/proxy_client.go
pbialon: Ok, I fixed it, now the output is as follows:

```
10:05:45 error could not find corresponding workspace
10:05:45 error Try using the --debug flag to see a more verbose output
10:05:45 fatal error importing a workspace: could not find corresponding workspace
```
So, we have the logs from the child command (could not find corresponding workspace) alongside with the proper error at the end.

89luca89 (approved):
LGTM

PR #696, opened 2023-09-17
add environment variable substitution in tar gz feature ids

aacebedo:
This PR add the substitution of environment variables in feature id of direct tar features
It enables using private gitlab registries to house feature easily

### DIFF
--- a/pkg/devcontainer/feature/features.go
+++ b/pkg/devcontainer/feature/features.go
@@ -239,6 +239,28 @@ func downloadFeatureFromURL(url string, destFile string, log log.Logger) error {
 		return errors.Wrap(err, "create feature folder")
 	}
 
+	log.Debugf("replace environment variables in URL")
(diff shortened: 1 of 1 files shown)


aacebedo:
corrected the lint errors

89luca89:
Hi @aacebedo 

Thanks for the PR!
We'll discuss internally about this, as for now this is diverging from the standard reference of devcontainer: https://containers.dev/implementors/json_reference/

We'll also follow the discussion in the upstream that you opened

aacebedo:
I did not receive an answer on the other thread yet but I think I can solve this by adding the http headers to use in the devpod customization field. This way it won't diverge from the spec. What do you think?

89luca89:
@aacebedo Yes if it's part of the customization then it can work :+1: so that if in the future upstream wants to support this, we'll not be out of spec

aacebedo:
I close this in favor of #715

PR #700, opened 2023-09-18
Small fixes to force-build

aacebedo:
Two fixes for the force-build flag that is very useful when building new features
First is the build is no triggered when the flag is used without a repository (so with skip-push)
Second is to force the redownload of the package when using archive on a https server

### DIFF
--- a/pkg/devcontainer/build.go
+++ b/pkg/devcontainer/build.go
@@ -103,7 +103,7 @@ func (r *runner) extendImage(ctx context.Context, parsedConfig *config.Substitut
 	}
 
 	// get extend image build info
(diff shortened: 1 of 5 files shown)


89luca89:
Hi @aacebedo can you post some example of devcontainer.json using this?

aacebedo:
Here it is:

{```

	"name": "test",
	"build" : {
                "dockerfile": "Dockerfile"
        },
        "features": {
                "https://www.gitlab.com/api/v4/projects/50844431/packages/generic/foo/0.0.0/devcontainer-feature-foo.tgz":{}
        }
}
```

and the Dockerfile
```
FROM ubuntu:latest

RUN apt update && apt install -y sudo
RUN useradd -g users -u 10000 devcontaineruser -m
USER devcontaineruser
```

89luca89:
Thanks @aacebedo LGTM!

PR #702, opened 2023-09-21
przemekbialon/eng 1594 plain devcontainerjson contains empty build key

pbialon:
Fix ENG-1594

- created a draft for the e2e test
- added code validating devcontainer.json contents

### DIFF
--- a/pkg/devcontainer/config/parse_test.go
+++ b/pkg/devcontainer/config/parse_test.go
@@ -1,0 +1,55 @@
+package config
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func TestSaveDevContainerJSON(t *testing.T) {
+	type args struct {
+		config *DevContainerConfig
+	}
+	tests := []struct {
+		name     string
+		args     args
+		wantErr  bool
+		wantJSON string
+	}{
+		{
+			name: "test omit build field in devcontainer.json",
+			args: args{
+				config: &DevContainerConfig{
+					ImageContainer: ImageContainer{
+						Image: "test",
+					},
+				},
+			},
+			wantErr:  false,
+			wantJSON: `{"image":"test"}`,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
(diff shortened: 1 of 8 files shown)


pkg/devcontainer/config/parse_test.go
```diff
@@ -0,0 +1,55 @@
...
+			wantJSON: `{"image":"test"}`,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			tmpDir, err := os.MkdirTemp("", "test-devcontainer")
+			if err != nil {
+				t.Fatalf("Failed to create temp dir: %v", err)
+			}
+			defer os.RemoveAll(tmpDir)
```
pascalbreuninger, line 38: nit: maybe can be run as the very first statement when the path is constructed with `os.TempDir()/test-devcontainer` as a replacement for the empty string? I'm a bit worried about interrupts here

pbialon: I replaced an empty string with `os.TempDir()`, but why would I run `os.RemoveAll(tmpDir)` before checking for error? If there is the error, then the directory was not created, so no need for removing it

pascalbreuninger: then we'll have to check if `os.IsExists`. If you the test run is interrupted for any reason, like ctrl + c, it will fail on subsequent runs because the directory still exists

pbialon: `TempDir()` creates temp directory with a random suffix, so when we run it again we will have a temp dir with different name.

pascalbreuninger: according to the docs it's returning the base temporary directory, like `/tmp`. Anyhow, this is just a nitpick anyway, let's get this merged 💪

pbialon: Hmm 🤔 
In `os.MkdirTemp` there is a description: 

> // MkdirTemp creates a new temporary directory in the directory dir
// and returns the pathname of the new directory.
// The new directory's name is generated by adding a random string to the end of pattern.

PR #704, opened 2023-09-22
fail fast if mounted filesystem for agent is noexec

pbialon:
Fix ENG-2043

Tested locally by manually switching `isNoExec()` return value for mac from `false` to `true`


pascalbreuninger (approved):
LGTM

PR #712, opened 2023-09-28
fix: setup-dotfiles should be ran as the default user

### LINKED ISSUE #655: Dotfile script is executed as root if external SSH config is used
Hi,

I'm not entirely sure what happened here, I just noticed weird behavior when playing with dotfiles and external SSH configs. I'll try to give as much information as possible.

**What happened?**  

Dotfile script was executed as `root` user if external SSH config is used (via `--ssh-config` option and `Include` in `~/.ssh/config`).

Dotfile script should be executed as `remoteUser`. As it normally does when not using `--ssh-config`. In this case, I expected the user to be `vscode`.

SSH Config `~/.ssh/config`:

```plain
Include devpod.conf
```

No custom `devcontainer.json` just DevPod command using prebuild container image:

```bash
devpod up mcr.microsoft.com/devcontainers/base:bookworm \
--dotfiles https://github.com/flokoe/dotfiles-devcontainer \
--dotfiles-script debug.sh \
--ide none \
--recreate \
--ssh-config ~/.ssh/devpod.conf
```

The output of my debug script just dumps information about the current environment. The user will be `root`

If the same command is executed without `--ssh-config ~/.ssh/devpod.conf` The user will be `vscode` as expected.

**Local Environment:**  
- DevPod Version: `v0.3.6`
- Operating System: Debian 12 (Bookworm)
- ARCH of the OS: `AMD64`

**DevPod Provider:**  
- Local/remote provider: `docker`
- Local Docker Version: `24.0.5`

I'm unsure if this behavior is due to the usage of `--ssh-config` or the include itself in the main SSH config. It's probably the include. As soon as a DevPod SSH config exists in `~/.ssh/config` it works again.

89luca89:
This way now we will default to the container user (usually vscode) and fallback to root otherwise

Fix #655 
Resolves ENG-1975

### DIFF
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -786,10 +786,10 @@ func setupDotfiles(
 		agentArguments = append(agentArguments, dotfilesScript)
 	}
 
-    remoteUser, err:= devssh.GetUser(client.Workspace())
-    if err != nil {
-        remoteUser = "root"
-    }
+	remoteUser, err := devssh.GetUser(client.Workspace())
+	if err != nil {
+		remoteUser = "root"
+	}
 
 	dotCmd := exec.Command(
 		execPath,
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -786,11 +786,18 @@ func setupDotfiles(
 		agentArguments = append(agentArguments, dotfilesScript)
 	}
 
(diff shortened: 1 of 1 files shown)


pascalbreuninger (approved):
LGTM

PR #715, opened 2023-09-30
Add http headers support when downloading features archives

aacebedo:
This feature enables to use devpod with servers needing an authentication.


aacebedo:
@89luca89 rebased and fixed the conflicts

aacebedo:
Ping?

pkg/devcontainer/feature/features.go
```diff
@@ -241,11 +242,22 @@ func downloadFeatureFromURL(url string, destFile string, log log.Logger) error {
 
 	// initiate download
 	log.Debugf("Download feature from %s", url)
-	resp, err := devpodhttp.GetHTTPClient().Get(url)
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return errors.Wrap(err, "make request")
+	}
+	for key, value := range httpHeaders {
+		req.Header.Set(key, value)
```
pascalbreuninger, line 250: Could you ensure we're not accidentally overwriting existing headers here please?

docs/pages/developing-in-workspaces/devcontainer-json.mdx
```diff
@@ -79,6 +79,27 @@ You can think of features as reusable Dockerfile parts that will be merged into
...
+
+```
+{
+  ...
+  "features": {
+    "https://example.com/foo_feature.tar.gz": []
+  },
+  "customizations": {
+    "devpod": {
+      "directDownloadHTTPHeaders": {
```
pascalbreuninger, line 94: `directDownloadHTTPHeaders` is ambiguous in this context imo. It's not immediately clear how this customization relates to features, could you rename it to something like `featureDownloadHTTPHeaders` or similar?

pascalbreuninger (changes requested):
Thanks for your contribution! 
Added two comments we should address before merging :) 

Additionally, would be great if you could add a test for this in `e2e/up`

pkg/devcontainer/feature/features.go
aacebedo: Do you mean I shall check for duplicates in the json file?
Because the req is created in the method, so it can't have any headers by default.

For the duplicate keys in the json it is going to be complicated. According to this https://github.com/golang/go/issues/48298, JSON unmarshaller is not able to deal with map containing duplicated keys.

However, the author states that this could be solved in json v2 package, so the solution would be to update the dependency in the devpod. If we still want to do this in devpod it means to manually parse the bytes in a specific unmarshall function for the features to detect duplicates just to display a warning. Not sure that it is worth the cost.

What do you think?

docs/pages/developing-in-workspaces/devcontainer-json.mdx
aacebedo: yep that's a better idea, I changed it.

aacebedo:
> Thanks for your contribution! Added two comments we should address before merging :)
> 
> Additionally, would be great if you could add a test for this in `e2e/up`

For the e2e/up test I think I'll need a zip package containing a dummy feature and stores it on a private package registry of the devpod repo.
Would it be ok for you ? And are you able to store this package in a private registry?

pkg/devcontainer/feature/features.go
pascalbreuninger: > Because the req is created in the method, so it can't have any headers by default.

Ah yeah right, dismiss what I said earlier.
In the case of duplicate keys in `devcontainer.json` it's a configuration error and doesn't need to be handled by DevPod imo

e2e/tests/up/up.go
```diff
@@ -33,7 +35,7 @@ var _ = DevPodDescribe("devpod up test suite", func() {
 		framework.ExpectNoError(err)
 
 		dockerHelper = &docker.DockerHelper{DockerCommand: "docker"}
-		composeHelper, err = compose.NewComposeHelper("", dockerHelper)
+		composeHelper, err = compose.NewComposeHelper("podman-compose", dockerHelper)
```
pascalbreuninger, line 38: could you remove that again pls?

e2e/tests/up/testdata/docker-features-http-headers/.devcontainer.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "Test hello",
+  "image": "mcr.microsoft.com/vscode/devcontainers/base:ubuntu",
+  "features": {
+    "@SERVER_URL@/devcontainer-feature-hello.tgz": {}
```
pascalbreuninger, line 5: super small nitpick: We use `#{}` to escape in `pkg/devcontainer/feature/extend.go`. Since it's related to this would be great to keep the escaping consistent

pascalbreuninger:
> > Thanks for your contribution! Added two comments we should address before merging :)
> > Additionally, would be great if you could add a test for this in `e2e/up`
> 
> ~For the e2e/up test I think I'll need a zip package containing a dummy feature and stores it on a private package registry of the devpod repo. Would it be ok for you ? And are you able to store this package in a private registry?~
> 
> Added a test and simulates the HTTP server during execution.

Great, thanks

e2e/tests/up/up.go
aacebedo: Fixed sorry for the mistake

e2e/tests/up/testdata/docker-features-http-headers/.devcontainer.json
aacebedo: done

pascalbreuninger (approved):
LGTM

edit: @aacebedo the test fails in CI

aacebedo:
I rebased the PR however locally it was working perfectly.
By taking a look to the test logs it seems it cannot find the resulting image and that's strange because I didn't change anything here.

Can you give it a look please? I do not have access to the GH actions infrastructure so it is hard to me to reproduce it.

pascalbreuninger:
@aacebedo finally worked out, thanks a lot for your contribution and patience!

aacebedo:
> @aacebedo finally worked out, thanks a lot for your contribution and patience!

Thanks to you too to have taken the time to review and merge this. It is going to be very useful to me.

aacebedo:
@pascalbreuninger just a small question, when do you plan to tag a new version of devpod?

pascalbreuninger:
@aacebedo We're waiting for #734 to be merged for the `v0.4.0` release but we can cut an alpha version right now

pascalbreuninger:
@aacebedo It's released in `v0.4.0` now

Issue #626, opened 2023-08-19
Using DevPod on Windows with WSL2


icebob:
**Is your feature request related to a problem?**  
I tested DevPod on my Windows 10 with Docker. Docker is installed in WSL2 in an Ubuntu 22.04 but the `docker.exe` executable in Windows which accesses the Docker Engine in WSL.

When I try to create a Workspace from a github repo (but same result with local folder, as well), DevPod tries to use the Windows path for Docker volume binding, which is incorrect inside WSL.

```
[11:30:07] debug Acquire workspace lock...
[11:30:07] debug Acquired workspace lock...
[11:30:07] info Creating devcontainer...
[11:30:07] debug Inject and run command: 'C:\Program Files\DevPod\devpod-cli.exe' agent workspace up --workspace-info 'H4sIAAAAAAAA/5yTX2+bMBTFv8p0n3EgkAbwW9V2arQ/qToyaZOlyrEviVeDkW1opyjffXJI02zdXvaGwffec+7vsIMnYx9dxwUC3YGSQKExGkWv0UIE/e9viMSa99qTaT4vURSZSEUymxYXhM9lJiCC2miJFihcUcZWDq1jbCFwbdaMTSQOnZGMCdN6fPaOsWM/xk4yHGPnAjolfG8RKGy97xyNY4udccob+5Oohm/QTTbKb/t179AeGrd+IkwTF2lalLMii8t1PZvWRUJyrHMynWJJSp4n5CKdizov5glmeZhkzaAO2nfQ8iaMlEY8HlSYzivTuvDpenn14eb+4e6yug3Hgev+7Oo+gsXny6tq8XVRfXuoFp9ulqsK6G6/30fQcLFVbVj0PgIl8WyS6bAdnDASQwtnejsC2Sh/f/J7toTR8sHnaVs/3Otz6CIs8iC7Ug06z5sOKKRJmpGkINOySkqaJTSZf4cINHd+5VD++8YRWXA6IgsT+AZbH2RqI7gGCt72GFbJ/fYYgTtrNpY3794rHdBe43AXEjAmgQitJvgcSqR5arXhcnX/8e82tak9cdt4rIwtauQOXfxSGA/JJJvMIAJ8RjHu+LFfo23RoxvPR0h096LwBFi1znMdLNRcO4SAS2i1fOX+5tc4APwD3Nuc3C6/VOPw/w7OLwAAAP//AQAA///dCuvbpAMAAA==' --debug
[...]

PR #729, opened 2023-10-08
add info on setting home dir

jerempy:
Closes: https://github.com/loft-sh/devpod/issues/626

After following down the chain found that there is the option to set the home path for mounts. Passing the flag allows to set the desired home directory, which is then later used for building mount paths.

This can allow someone to set or change where the mounts should be happening by specifying the homedir either as flag or env var

### DIFF
--- a/docs/pages/developing-in-workspaces/create-a-workspace.mdx
+++ b/docs/pages/developing-in-workspaces/create-a-workspace.mdx
@@ -33,6 +33,16 @@ Then press `Create Workspace` to launch the workspace.
 Under the hood, the Desktop Application will call the CLI command `devpod up REPOSITORY`
 :::
 
+:::info Note
+You can set the location of your devpod home by passing the `--devpod-home={home_path}` flag, 
+or by setting the env var `DEVPOD_HOME` to your desired home directory.
+
+This can be useful if you are having trouble with a workspace trying to mount to a windows location when it should be mounting
+to a path inside the WSL VM.
+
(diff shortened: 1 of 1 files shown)


89luca89 (approved):
LGTM

docs/pages/developing-in-workspaces/create-a-workspace.mdx
```diff
@@ -33,6 +33,16 @@ Then press `Create Workspace` to launch the workspace.
...
 :::
 
+:::info Note
+You can set the location of your devpod home by passing the `--devpod-home={home_path}` flag, 
+or by setting the env var `DEVPOD_HOME` to your desired home directory.
+
+This can be useful if you are having trouble with a workspace trying to mount to a windows location when it should be mounting
+to a path inside the WSL VM.
+
+For example: setting `devpod-home=/mnt/c/Users/MyUser/` will result in a workspace mount path of something like `/mnt/c/Users/MyUser/.devpod/agent/contexts/default/workspaces/...`
```
pascalbreuninger, line 43: the `agent` directory only appears here if you're using a local provider, can you remove that pls? This might confuse people if they aren't able to find it on their machine because they are using a remote provider

pascalbreuninger (changes requested):
`/mnt/c/Users/MyUser/.devpod/agent/contexts/default/workspaces/...`
-> `/mnt/c/Users/MyUser/.devpod/contexts/default/workspaces/...`

the `agent` directory only appears here if you're using a local provider, can you remove that pls? This might confuse people if they aren't able to find it on their machine because they are using a remote provider

pascalbreuninger (approved):

pascalbreuninger:
> ok thanks! i updated this

great, thank you :bow:

PR #734, opened 2023-10-11
Refactor DevPod up command to allow SSH agent forwarding

### LINKED ISSUE #673: SSH agent not being forwarded during `postCreateCommand`
**What happened?**  
I've realized that the SSH agent is not being forwarded during `postCreateCommand` because the SSH connection is not being established yet.

I expect `postCreateCommand` to have access to the forwarded agent (same as when running locally), because it can be needed to pull from private repositories during that (npm, pip install, etc.).

1. Setup any private Git repo to which you have access with your local SSH keys.
2. Have a devcontainer.json that tries to checkout the same private Git repo over SSH in `postCreateCommand`.
2. Notice how it cannot clone it.

**Local Environment:**  
- DevPod Version: 0.3.3
- Operating System: mac
- ARCH of the OS: ARM64

**DevPod Provider:**  
- Cloud Provider: AWS

neogopher:
Closes ENG-2020
Closes ENG-2006
Closes #673

### DIFF
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -19,6 +19,7 @@ import (
 	"github.com/loft-sh/devpod/pkg/command"
 	"github.com/loft-sh/devpod/pkg/config"
 	config2 "github.com/loft-sh/devpod/pkg/devcontainer/config"
+	"github.com/loft-sh/devpod/pkg/devcontainer/sshtunnel"
 	"github.com/loft-sh/devpod/pkg/ide/fleet"
 	"github.com/loft-sh/devpod/pkg/ide/jetbrains"
 	"github.com/loft-sh/devpod/pkg/ide/jupyter"
@@ -36,7 +37,6 @@ import (
 	"github.com/skratchdot/open-golang/open"
 	"github.com/spf13/cobra"
 	"golang.org/x/crypto/ssh"
-	gosshagent "golang.org/x/crypto/ssh/agent"
 )
 
 // UpCmd holds the up cmd flags
@@ -351,15 +351,17 @@ func (cmd *UpCmd) devPodUpMachine(
 		return nil, err
 	}
 
-	// ssh tunnel
-	sshCmd := fmt.Sprintf("'%s' helper ssh-server --stdio", client.AgentPath())
-	if log.GetLevel() == logrus.DebugLevel {
-		sshCmd += " --debug"
-	}
-
 	// create container etc.
 	log.Infof("Creating devcontainer...")
 	defer log.Debugf("Done creating devcontainer")
+
+	// ssh tunnel command
+	sshTunnelCmd := fmt.Sprintf("'%s' helper ssh-server --stdio", client.AgentPath())
+	if log.GetLevel() == logrus.DebugLevel {
+		sshTunnelCmd += " --debug"
+	}
+
+	// create agent command
 	agentCommand := fmt.Sprintf(
 		"'%s' agent workspace up --workspace-info '%s'",
 		client.AgentPath(),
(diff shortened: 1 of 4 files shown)


89luca89 (approved):
LGTM

Maybe we can find a way to e2e test this?

pkg/devcontainer/setup/setup.go
```diff
@@ -206,6 +212,20 @@ func PatchEtcEnvironment(mergedConfig *config.MergedDevContainerConfig, log log.
 	return nil
 }
 
+func ChownAgentSock(setupInfo *config.Result, log log.Logger) error {
+	user := config.GetRemoteUser(setupInfo)
+
+	agentSockFile := os.Getenv("SSH_AUTH_SOCK")
+	if agentSockFile != "" {
+		err := copy2.ChownR(filepath.Dir(agentSockFile), user)
```
pascalbreuninger, line 220: why do we need to `chown` the whole directory here?

pkg/devcontainer/setup/setup.go
```diff
@@ -206,6 +212,20 @@ func PatchEtcEnvironment(mergedConfig *config.MergedDevContainerConfig, log log.
 	return nil
 }
 
+func ChownAgentSock(setupInfo *config.Result, log log.Logger) error {
```
pascalbreuninger, line 215: we never return an error in the function body, is this on accident?

cmd/up.go
```diff
@@ -401,19 +408,82 @@ func (cmd *UpCmd) devPodUpMachine(
...
+		if identityAgent == "" {
+			identityAgent = os.Getenv("SSH_AUTH_SOCK")
+		}
+
+		if identityAgent != "" {
+			err = gosshagent.ForwardToRemote(sshClient, identityAgent)
+			if err != nil {
+				errChan <- errors.Wrap(err, "forward agent")
+			}
+			err = gosshagent.RequestAgentForwarding(sess)
```
pascalbreuninger, line 470: just leaving this as a marker: @neogopher and me had a discussion around why this works on the session even though we create a new one in `devssh.Run`, he investigated a potential bug in the ssh library dependency and I'll also check it out this week to be sure

cmd/up.go
```diff
@@ -350,16 +351,22 @@ func (cmd *UpCmd) devPodUpMachine(
...
 	}
 
+	// ssh tunnel
+	sshCmd := fmt.Sprintf("'%s' helper ssh-server --stdio", client.AgentPath())
+	if log.GetLevel() == logrus.DebugLevel {
+		sshCmd += " --debug"
+	}
+
 	// create container etc.
 	log.Infof("Creating devcontainer...")
```
pascalbreuninger, line 361: can you create the agent command where we actually execute it or is there a reason to do it here?

cmd/up.go
```diff
@@ -401,19 +408,82 @@ func (cmd *UpCmd) devPodUpMachine(
...
-		if err != nil {
+		if err != nil && !errors.Is(err, context.Canceled) && !strings.Contains(err.Error(), "signal: ") {
 			errChan <- fmt.Errorf("executing agent command: %w", err)
 		} else {
 			errChan <- nil
 		}
 	}()
 
+	// create pipes
+	stdoutReader2, stdoutWriter2, err := os.Pipe()
```
pascalbreuninger, line 425: could you rename these and `stdinReader/Writer2` below pls? It's currently not clear what they are being used for without scrolling down

pkg/devcontainer/setup.go
```diff
@@ -95,28 +104,90 @@ func (r *runner) setupContainer(
...
+		return nil, err
+	}
+	stdinReader2, stdinWriter2, err := os.Pipe()
+	if err != nil {
+		return nil, err
+	}
+	defer stdoutWriter2.Close()
+	defer stdinWriter2.Close()
+
+	go func() {
```
pascalbreuninger, line 136: we're doing this whole setup once on the machine and once in the container. Do you think it's worth refactoring into a separate pkg or function?

pascalbreuninger (changes requested):

cmd/up.go
```diff
@@ -401,19 +408,82 @@ func (cmd *UpCmd) devPodUpMachine(
...
+		defer sess.Close()
+
+		log.Debugf("SSH session created")
+
+		var identityAgent string
+		if identityAgent == "" {
+			identityAgent = os.Getenv("SSH_AUTH_SOCK")
+		}
+
+		if identityAgent != "" {
```
FabianKramm, line 465: Can we simplify this to:
```
identityAgent := os.Getenv("SSH_AUTH_SOCK")
```

pkg/devcontainer/setup/setup.go
neogopher: I had tried that initially.
So the way the agent sock parent directories are created is like below. Notice how root is the owner of the directory and has full permissions while group and others have no permission whatsoever.
```
$ ls -l /tmp/
drwx------ 2 root root  4096 Oct 26 07:38 auth-agent3289311155
drwx------ 2 root root  4096 Oct 26 07:04 auth-agent961926320
```

Now, if I just change the owner of the sock file, I get permissions on the file, but still no access on the parent directory. Due to this, we get _: Permission denied_ as it can't traverse the directory chain.

There are 2 options to make this work:
1. Give the read and execute permissions to the _others_ group on this directory and the sock file.
2. Change the owner of the directory and the file to the remoteUser.

I have used option 2.

cmd/up.go
neogopher: Done.

cmd/up.go
neogopher: Done.

cmd/up.go
neogopher: Done

pkg/devcontainer/setup/setup.go
pascalbreuninger: I see, thanks for the explanation

pascalbreuninger (approved):
LGTM

PR #738, opened 2023-10-12
Set SSH working directory

### LINKED ISSUE #730: SSH starting directory should be workspace directory
**What happened?**  

SSH into a workspace starts at user home directory (perhaps Docker image workdir).

Current directory should automatically be the workspace directory.

1. Create a workspace for Node.js
2. SSH into the workspace with `devpod ssh`
3. `pwd` will show `/home/node` as opposed to `/workspaces/<name>`

**Local Environment:**  
- DevPod Version: v0.3.7
- Operating System: linux & mac
- ARCH of the OS: AMD64 & ARM64 

**DevPod Provider:**  
- Localprovider: docker

frangio:
Fixes #730

I didn't go as far as adding a `--workdir` argument to `devpod ssh` as suggested in https://github.com/loft-sh/devpod/issues/730#issuecomment-1759072845 because it isn't strictly necessary.

Looking forward to feedback.

Should this be tested in any way?

### DIFF
--- a/cmd/ssh.go
+++ b/cmd/ssh.go
@@ -5,9 +5,9 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"sync"
 	"time"
-	"path/filepath"
 
 	"github.com/loft-sh/devpod/cmd/flags"
 	"github.com/loft-sh/devpod/cmd/machine"
--- a/cmd/ssh.go
+++ b/cmd/ssh.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"sync"
 	"time"
+	"path/filepath"
 
 	"github.com/loft-sh/devpod/cmd/flags"
 	"github.com/loft-sh/devpod/cmd/machine"
@@ -124,7 +125,7 @@ func (cmd *SSHCmd) startProxyTunnel(ctx context.Context, devPodConfig *config.Co
 			})
 		},
 		func(ctx context.Context, containerClient *ssh.Client) error {
-			return cmd.startTunnel(ctx, devPodConfig, containerClient, client.WorkspaceConfig().IDE.Name, log)
+			return cmd.startTunnel(ctx, devPodConfig, containerClient, client.Workspace(), client.WorkspaceConfig().IDE.Name, log)
 		},
 	)
 }
@@ -191,7 +192,7 @@ func (cmd *SSHCmd) jumpContainer(ctx context.Context, devPodConfig *config.Confi
 		unlockOnce.Do(client.Unlock)
 
 		// start ssh tunnel
-		return cmd.startTunnel(ctx, devPodConfig, containerClient, client.WorkspaceConfig().IDE.Name, log)
+		return cmd.startTunnel(ctx, devPodConfig, containerClient, client.Workspace(), client.WorkspaceConfig().IDE.Name, log)
 	})
 }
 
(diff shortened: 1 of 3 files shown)


cmd/ssh.go
```diff
@@ -243,7 +244,7 @@ func (cmd *SSHCmd) startTunnel(ctx context.Context, devPodConfig *config.Config,
 	defer writer.Close()
 
 	log.Debugf("Run outer container tunnel")
-	command := fmt.Sprintf("'%s' helper ssh-server --track-activity --stdio", agent.ContainerDevPodHelperLocation)
+	command := fmt.Sprintf("'%s' helper ssh-server --track-activity --stdio --workdir '%s'", agent.ContainerDevPodHelperLocation, filepath.Join("workspaces", workspaceName))
```
frangio, line 247: Can workspace names have problematic symbols that may need to be escaped?

frangio, line 247: On second thought `filepath.Join("workspaces", workspaceName)` is probably not right... What would be the right way to get the directory?

frangio: `ContainerWorkspaceFolder` is the only thing I could find but it seems available at a very different point in the lifecycle.

https://github.com/loft-sh/devpod/blob/82e870c14b8834d3b3fedee3e1e94364b836e4a9/pkg/devcontainer/config/substitute.go#L26

pascalbreuninger: You will need to make workspace dir absolute `/workspaces`. 
I guess it's fine as a default, when users set `workspaceFolder` in the `devcontainer.json`, it won't be able to find this path and fall back to the default

pkg/ssh/server/ssh.go
```diff
@@ -328,10 +330,13 @@ func (s *Server) getCommand(sess ssh.Session, isPty bool) *exec.Cmd {
...
-	_, err := os.Stat(home)
+	workdir := s.workdir
+	if workdir == "" {
+		workdir, _ = command.GetHome(user)
+	}
+	_, err := os.Stat(workdir)
 	if err == nil {
-		cmd.Dir = home
+		cmd.Dir = workdir
 	}
```
pascalbreuninger, line 340: we want to fall back to the user home directory here if the path set by `s.workdir` can't be found. Right now if `os.Stat` fails we end up in `/root` 😅

pascalbreuninger (changes requested):

cmd/ssh.go
pascalbreuninger: yeah but we don't have a parsed `devcontainer.json` if I'm not mistaken

frangio: That seems correct to me there doesn't seem to be a parsed devcontainer.json, only a devpod config.

pascalbreuninger (approved):
LGTM, thanks for your contribution 🙇

frangio:
Should be ready now. Thanks!

Do you have an idea when this might be released?

frangio:
Do you think the E2E test failures are caused by these changes?

pascalbreuninger:
> Do you think the E2E test failures are caused by these changes?

Yep, will have a look tomorrow

FabianKramm:
We need to check if Fleet is still working after this check as they expect the SSH root to be in HOME as thats the default for SSH I think.

pascalbreuninger:
@FabianKramm fleet stilled worked for me and tests are passing now

pascalbreuninger:
@frangio It's released in `v0.4.0` now

PR #771, opened 2023-10-23
fix: resolve regression in docker auto-installation logic

89luca89:
Resolve logic error in docker auto-instalation routine

Resolves ENG-2287


pascalbreuninger (approved):
LGTM

PR #779, opened 2023-10-27
fix: fleet should start if already installed

89luca89:
Simple fix for fleet not starting after a first setup

Resolves ENG-2218


pascalbreuninger (approved):
LGTM

PR #780, opened 2023-10-27
fix: improve error logging across commands and postcreate commands

### LINKED ISSUE #775: Errors in postCreateCommand are not logged
Errors in postCreateCommand are not logged, even in debug.

The repo has an invalid command here:
https://github.com/behrica/try-langchain-opencity-invalid/blob/76559acab97050985ced7766a799784a62fa73a0/.devcontainer/devcontainer.json#L39

but there is no error in here, except the "21:55:34 fatal executing agent command: agent error: No root required, because neither docker nor agent daemon needs to be installed"

```
devpod up --provider ssh  --ide none  --recreate https://github.com/behrica/try-langchain-opencity-invalid.git --debug
```

89luca89:
Generally improve STDERR logging by using the error level instead of info level
Add more messages to postcreate command execution

Giving now a more clear and first-glanceable output

image

Fix #775 
Resolves ENG-2298

### DIFF
--- a/pkg/devcontainer/setup/setup.go
+++ b/pkg/devcontainer/setup/setup.go
@@ -332,7 +332,7 @@ func runPostCreateCommand(commands []types.LifecycleHook, user, dir string, remo
 			log.Debugf("Executing command %s: %s...", k, cmd.Args)
 			err := cmd.Run()
 			if err != nil {
-				log.Debugf("Failed running postcreate command %s: %v", cmd.Args, err)
+				log.Debugf("Failed running postCreateCommand lifecycle script %s: %v", cmd.Args, err)
 				return fmt.Errorf("failed to run: %s, error: %w", strings.Join(c, " "), err)
 			}
 			log.Donef("Successfully ran command %s: %s", k, strings.Join(c, " "))
--- a/pkg/devcontainer/setup/setup.go
+++ b/pkg/devcontainer/setup/setup.go
@@ -2,6 +2,7 @@ package setup
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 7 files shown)


pkg/devcontainer/sshtunnel/sshtunnel.go
```diff
@@ -103,7 +103,7 @@ func ExecuteCommand(ctx context.Context, client client2.WorkspaceClient, agentIn
 			}
 		}
 
-		writer := log.Writer(logrus.InfoLevel, false)
+		writer := log.Writer(logrus.ErrorLevel, false)
```
pascalbreuninger, line 106: We shouldn't change this to `ErrorLevel` here as all log output will be piped over stderr from the agent command because stdout is used for the connection. 
I checked it locally and the post create command output still shows up as `error` when this is info. 
So we could either revert it or implement a proper check on this writer and see which log level the agent command sent originally

pkg/devcontainer/setup/setup.go
```diff
@@ -325,11 +328,12 @@ func runPostCreateCommand(commands []types.LifecycleHook, user, dir string, remo
...
 			cmd.Env = append(cmd.Env, remoteEnvArr...)
 			cmd.Stdout = writer
-			cmd.Stderr = writer
+			cmd.Stderr = errwriter
+			log.Debugf("Executing command %s: %s...", k, cmd.Args)
 			err := cmd.Run()
 			if err != nil {
-				log.Errorf("Failed running command %s: %v", k, err)
-				return err
+				log.Debugf("Failed running postcreate command %s: %v", cmd.Args, err)
```
pascalbreuninger, line 335: `postcreate` might be a little confusing here. I know we call the function that way but the spec calls these ["lifecycle script"](https://containers.dev/implementors/json_reference/#lifecycle-scripts)

pascalbreuninger (changes requested):
Pretty nice log output now overall, had 1 nit and 1 required change

pkg/devcontainer/sshtunnel/sshtunnel.go
89luca89: yes but this writer is specifically used for the STDERR in the devssh.Run function
And usually if something is in stderr, it should be an error

Probably just a nitpick on my side, I can revert if this creates problems let me know

Issue #791, opened 2023-10-31
In README.md Twitter's new name is not updated to 𝕏.


BandhiyaHardik:
Current Behaviour:
![image](https://github.com/loft-sh/devpod/assets/110784317/c1d42385-7a7b-4c81-9850-99553ca31d90)

Expected Behaviour:
![image](https://github.com/loft-sh/devpod/assets/110784317/6216edd3-0452-4e6c-aeed-d408c277ebab)

PR #792, opened 2023-10-31
chore: In README.md Twitter's new name updated to 𝕏. 

BandhiyaHardik:
Twitter is now known as 𝕏.

So we should reflect it in our repository.

This fixes https://github.com/loft-sh/devpod/issues/791

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
   </picture>
 </a>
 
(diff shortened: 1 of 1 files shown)


pascalbreuninger (approved):
Thanks for the PR

PR #799, opened 2023-11-07
tests: parallelize tests using labels and matrix

89luca89:
Improve e2e tests by running them in separate jobs, so that they can fail earlier and run in parallel

Resolves ENG-2326


FabianKramm:
@89luca89 the problem I see here is that a couple of tests are super fast, but up takes 18m alone, can we split up the `up` test and merge a couple of fast ones together?

89luca89:
> @89luca89 the problem I see here is that a couple of tests are super fast, but up takes 18m alone, can we split up the `up` test and merge a couple of fast ones together?

Sure, I'll check a logic to split them, so we also have some granularity with job retrying

89luca89:
> @89luca89 the problem I see here is that a couple of tests are super fast, but up takes 18m alone, can we split up the `up` test and merge a couple of fast ones together?

Done

pascalbreuninger (approved):
LGTM

PR #803, opened 2023-11-10
fix: launch gpg-forwarding during UP if flag is specified

89luca89:
This fixes a behavioural bug,
if we specify the `--gpg-agent-forwarding` during `up`, we expect it to take immediate effect when using our IDE

This fixes the inconsistency

Resolves ENG-1900

### DIFF
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -184,13 +184,6 @@ func (cmd *UpCmd) Run(
 		return err
 	}
 
-	if cmd.GPGAgentForwarding {
-		err = performGpgForwarding(client, log)
-		if err != nil {
-			return err
-		}
-	}
-
 	// open ide
 	if cmd.OpenIDE {
 		ideConfig := client.WorkspaceConfig().IDE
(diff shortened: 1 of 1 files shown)


FabianKramm:
@89luca89 thanks for the PR! I'm not sure why this is needed, isn't gpg forwarding only useful during ssh forwarding?

89luca89:
> @89luca89 thanks for the PR! I'm not sure why this is needed, isn't gpg forwarding only useful during ssh forwarding?

More like a behaviour uniformity
When I say `up --gpg-agent-forwarding` I expect to have gpg agent forwarding immediately, not at the next ssh command
Now this is not a problem with IDEs based on ssh (vscode, jetbrains) but for other non-ssh based IDEs, this won't be the case

For example with fleet or with openvscode (web)

Same goes for forwarded ssh agent and so on, this ensures the behaviour is the same on all IDEs

FabianKramm:
@89luca89 can we move this logic then only to those ides? Also why is that required for fleet?

89luca89:
> @89luca89 can we move this logic then only to those ides? Also why is that required for fleet?

Done

pascalbreuninger (approved):
LGTM

PR #814, opened 2023-11-22
tests: add Windows E2E testing

89luca89:
This PR will add a test step for windows

Not all e2e tests are yet compatible with Windows, but it's a starting point

Resolves ENG-2368

### DIFF
--- a/e2e/framework/util.go
+++ b/e2e/framework/util.go
@@ -11,7 +11,7 @@ import (
 	"github.com/otiai10/copy"
 )
 
-func GetTiemout() time.Duration {
+func GetTimeout() time.Duration {
 	if runtime.GOOS == "windows" {
 		return 300 * time.Second
 	}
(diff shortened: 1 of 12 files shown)


e2e/framework/util.go
```diff
@@ -1,12 +1,24 @@
...
 	"os"
 	"path/filepath"
+	"runtime"
+	"strings"
+	"time"
 
 	"github.com/otiai10/copy"
 )
 
+func GetTiemout() time.Duration {
```
pascalbreuninger, line 14:
```suggestion
func GetTimeout() time.Duration {
```
nit: typo

pascalbreuninger (changes requested):

e2e/framework/util.go
89luca89: Fixed thanks!

pascalbreuninger (approved):
LGTM

PR #817, opened 2023-11-25
Fix ssh-config file management

aacebedo:
I add some issues using devpod on NixOS as on this system the file `~/.ssh/config` can be read-only. So this PR does the following:
- Add an option to context in order to be able to set the default path of the sshconfig file to anything. This enables, on systems where the `.ssh/config` is not accessible, to be set by default to something else.
- Fix an error when deleting containers when the `ssh-config` option has been given.


aacebedo:
@pascalbreuninger would you have some time to check this out please?

pkg/ssh/config.go
```diff
@@ -139,13 +139,16 @@ func writeSSHConfig(path, content string, log log.Logger) error {
...
-		return "", errors.Wrap(err, "get home dir")
+func GetDefaultSSHConfigPath() (string, error) {
+	if config.ContextOptionSSHConfigPath == "" {
+		homeDir, err := homedir.Dir()
+		if err != nil {
+			return "", errors.Wrap(err, "get home dir")
+		}
+		return filepath.Join(homeDir, ".ssh", "config"), nil
+	} else {
+		return config.ContextOptionSSHConfigPath, nil
```
pascalbreuninger, line 150: If I'm not mistaken `homedir.Dir()` returns an absolutes path. Let's do the same for the `SSHConfigPath` in case it's relative

pkg/ssh/config.go
```diff
@@ -86,14 +87,10 @@ func addHost(path, host, user, context, workspace, command string, gpgagent bool
 	return strings.Join(newLines, "\n"), nil
 }
 
-func GetUser(workspace string) (string, error) {
-	sshConfigPath, err := getSSHConfig()
-	if err != nil {
-		return "", err
-	}
-
+func GetUser(workspace provider.Workspace) (string, error) {
```
pascalbreuninger, line 90: nit: could you do `workspaceID, sshConfigPath string) like in the other functions please?

cmd/ssh.go
```diff
@@ -89,6 +89,12 @@ func NewSSHCmd(flags *flags.GlobalFlags) *cobra.Command {
 
 // Run runs the command logic
 func (cmd *SSHCmd) Run(ctx context.Context, devPodConfig *config.Config, client client2.BaseWorkspaceClient, log log.Logger) error {
+	// check if regular workspace client
```
pascalbreuninger, line 92: Is there a reason for moving this up?

pascalbreuninger (changes requested):
Thanks for your PR! I've noted a couple of smaller things to change and then re-run the e2e tests

pkg/ssh/config.go
aacebedo: fixed

cmd/ssh.go
aacebedo: mistakes from a previous attempt. fixed

aacebedo:
> Thanks for your PR! I've noted a couple of smaller things to change and then re-run the e2e tests

I have trouble to run e2e tests because I don't have the exact same setup and I am runnin

> Thanks for your PR! I've noted a couple of smaller things to change and then re-run the e2e tests

I tried to fix the e2e tests but without success. It looks like a side effect but for the moment I am unable to find it

aacebedo:
> > Thanks for your PR! I've noted a couple of smaller things to change and then re-run the e2e tests
> 
> I have trouble to run e2e tests because I don't have the exact same setup and I am runnin
> 
> > Thanks for your PR! I've noted a couple of smaller things to change and then re-run the e2e tests
> 
> I tried to fix the e2e tests but without success. It looks like a side effect but for the moment I am unable to find it

found a first error. Made a mistake in `GetDefaultSSHConfigPath`

aacebedo:
> > > Thanks for your PR! I've noted a couple of smaller things to change and then re-run the e2e tests
> > 
> > 
> > I have trouble to run e2e tests because I don't have the exact same setup and I am runnin
> > > Thanks for your PR! I've noted a couple of smaller things to change and then re-run the e2e tests
> > 
> > 
> > I tried to fix the e2e tests but without success. It looks like a side effect but for the moment I am unable to find it
> 
> found a first error. Made a mistake in `GetDefaultSSHConfigPath`

@pascalbreuninger can you take a look aqain. I had to redesign parts of the solution to fix the bug. I think the e2e tests are working now (I have checked the one I could).

aacebedo:
ping

pascalbreuninger:
@aacebedo I haven't forgotten about this PR, I first wanted to check if there's a simpler solution without passing the DevPod Config everywhere. Will get back to you soon :)

pascalbreuninger:
Can you check if these changes work for you as well please? @aacebedo  https://github.com/loft-sh/devpod/pull/835

pascalbreuninger:
great, I'll push the changes to this PR then so you're the PR owner :)

aacebedo:
> great, I'll push the changes to this PR then so you're the PR owner :)
Thanks !

89luca89 (approved):
LGTM

pascalbreuninger (approved):

PR #830, opened 2023-12-04
tests: add windows self hosted runner, add win tests in pipeline

89luca89:
Resolves ENG-2441

Add e2e tests using self-hosted windows runner

We do 2 rounds of tests, a faster/smaller subset for PRs, and full suite (compatible with windows) on pre-releases
Also we run the regular e2e tests on pre-releases too


pascalbreuninger (approved):
LGTM

PR #837, opened 2023-12-11
fix(cd): Exclude xdg-open and glib-2.0 schemas from Linux tar.gz

EyeCantCU:
While packaging this as an RPM, xdg-open and glib-2.0 were found in the Linux tar.gz archive.

Looking into this further, I extracted the last built RPM and saw that neither xdg-open nor the glib-2.0 schemas were there.

This excludes both from the release workflow.


89luca89:
Thanks @EyeCantCU LGTM!

PR #843, opened 2023-12-13
fix: forward git user.signingkey configuration when using gpg forwarding

### LINKED ISSUE #824: gpg forwarding bug - git user.signingkey configuration not transferred
**What happened?**  
I tried out the new GPG agent forwarding feature and when I tried to commit a change, I got the following error:

```
error: gpg failed to sign the data:
gpg: skipped <redacted>: No secret key
[GNUPG:] INV_SGNR 9 <redacted>
[GNUPG:] FAILURE sign 17
gpg: signing failed: No secret key
```

I expected my git commit to be signed successfully. Instead I had to first run `git config --global user.signingkey <redacted>` within the devcontainer environment to mirror my local setup.

A full reproduction requires setting the host up to use commit signing and then trying to do so from within the devcontainer. However, a much quicker reproduction is simply adding a dummy value for the host's global Git user.signingkey config and observing that it is not copied over automatically to the devcontainer's global git config alongside the user.name and user.email values.

I believe user.signingkey needs to be copied over in pkg/gitcredentials/gitcredentials.go.

**Local Environment:**  
- DevPod Version: v0.4.2
- Operating System: mac
- ARCH of the OS: AMD64

**DevPod Provider:**  
- Local/remote provider: ssh

That's it!

89luca89:
This PR will add `git config user.signingKey` detected from the host's gpg into the devcontainer, if we're using gpg agent forwarding.

Fix #824 
Resolves ENG-2498

### DIFF
--- a/pkg/gpg/gpg_forwarding.go
+++ b/pkg/gpg/gpg_forwarding.go
@@ -21,6 +21,7 @@ type GPGConf struct {
 	PublicKey  []byte
 	OwnerTrust []byte
 	SocketPath string
+	GitKey     string
 }
 
 func IsGpgTunnelRunning(
@@ -86,6 +87,19 @@ func (g *GPGConf) ImportOwnerTrust() error {
 	}()
 
 	return gpgOwnerTrustCmd.Run()
+}
+
+func (g *GPGConf) SetupGpgGitKey() error {
+	if g.GitKey != "" {
+		gitConfigCmd := exec.Command("git", []string{"config", "--global", "user.signingKey", g.GitKey}...)
+
+		out, err := gitConfigCmd.Output()
+		if err != nil {
+			return fmt.Errorf("git signkey: %s", string(out))
+		}
+	}
+
+	return nil
 }
 
 func (g *GPGConf) SetupGpgConf() error {
--- a/cmd/agent/workspace/setup_gpg.go
+++ b/cmd/agent/workspace/setup_gpg.go
@@ -17,6 +17,7 @@ type SetupGPGCmd struct {
 	PublicKey  string
 	OwnerTrust string
 	SocketPath string
+	GitKey     string
 }
 
 // NewSetupGPGCmd creates a new command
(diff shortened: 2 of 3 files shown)


pkg/gpg/gpg_forwarding.go
```diff
@@ -88,6 +89,19 @@ func (g *GPGConf) ImportOwnerTrust() error {
 	return gpgOwnerTrustCmd.Run()
 }
 
+func (g *GPGConf) SetupGpgGitKey() error {
+	if g.GitKey != "" {
+		gitConfigCmd := exec.Command("git", []string{"config", "--global", "user.signingKey", g.GitKey}...)
+
+		out, err := gitConfigCmd.Output()
+		if err != nil {
+			return fmt.Errorf("git signkey: %s", string(out))
```
pascalbreuninger, line 98: nit: `errors.Wrap` for potential stacktrace?

pkg/gpg/gpg_forwarding.go
```diff
@@ -88,6 +89,19 @@ func (g *GPGConf) ImportOwnerTrust() error {
 	return gpgOwnerTrustCmd.Run()
 }
 
+func (g *GPGConf) SetupGpgGitKey() error {
+	if g.GitKey != "" {
+		gitConfigCmd := exec.Command("git", []string{"config", "--global", "user.signingKey", g.GitKey}...)
```
pascalbreuninger, line 94: another small nit: can this go into the `gitcredentials` package? For the sole reason of keeping one place where we set the git config

pascalbreuninger (changes requested):

pkg/gpg/gpg_forwarding.go
89luca89: Done :+1:

pkg/gpg/gpg_forwarding.go
89luca89: Done :+1:

PR #868, opened 2024-01-12
fix: fail early during inject on noexec filesystems

### LINKED ISSUE #692: SSH provider should fail early on noexec filesystem
The default agent path of /tmp/devpod/agent could be mounted as noexec on some systems. When trying to start a workspace the agent will be downloaded and then fail to run. This repeats multiple times until failure with no way to cancel. No indication of the problem is given unless debug is enabled.

Detect if attempting to start the agent on a noexec mounted filesystem and fail early with an error message. Do not attempt to try again.

**Additional context**

89luca89:
Resolves ENG-2043
Fix #692 

Fail early during agent injection on noexec detection

### DIFF
--- a/pkg/inject/inject.sh
+++ b/pkg/inject/inject.sh
@@ -35,6 +35,22 @@ is_arm() {
   esac
 }
 
+# Detect if install_dir is noexec
+# We use this method instead of findmnt, as that command might not be present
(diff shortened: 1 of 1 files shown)


pascalbreuninger (approved):
LGTM!

PR #876, opened 2024-01-25
Update install.mdx to prorperly support ubuntu

souhaiebtar:
for ubuntu installing fuse make the system unusable, also it's not needed,


pascalbreuninger:
@souhaiebtar thanks for you PR. Would be great if you could add a sidenote specifically for ubuntu, might break not work distros if we just remove fuse

pascalbreuninger:
closing due to inactivity, feel free to reopen

PR #877, opened 2024-01-28
added opentelekomcloud as a community provider

akyriako:
Added support for Open Telekom Cloud, the OpenStack-based public cloud offering of the German leading IT services provider T-Systems International GmbH. 

That is a community project, not an officially supported one by T-Systems International GmbH.


pascalbreuninger:
Let's go, thanks for the PR @akyriako! Telekom cloud seems to be catching up

pascalbreuninger (approved):

PR #878, opened 2024-01-31
Fix/inactivity timeout

### LINKED ISSUE #776: Does inactivity timeout with ssh provider work?
**What happened?**  

I've set inactivity timeout for my ssh provider to 10m, but no container gets stopped after this time.

I'm expecting that a container is shutdown after the specified time.

- Using VSCode as IDE
- ssh-connection to a Ubuntu-VM

**Local Environment:**  
- DevPod Version: 0.3.7
- Operating System: mac
- ARCH of the OS: AMD64

**DevPod Provider:**  
- Local/remote provider: ssh

### LINKED ISSUE #795: Inactivity timeout stopping workspace although actively working in workspace
**What happened?**  
I’ve the inactivity timeout set to e.g. 6 hrs. While I was actively working in the workspace it got stopped. Tried with different values and behavior is consistent

My *`devcontainer.json`*:
```
{
    "name": "...",
    ...
}
```

**Local Environment:**  
- DevPod Version: 0.4.0
- Operating System: Mac
- ARCH of the OS:  ARM64 

**DevPod Provider:**  
- Cloud Provider: none
- Kubernetes Provider: none
- Local/remote provider: local docker desktop
- Custom provider: 

**Anything else we need to know?**  using fleet 1.25 as ide

89luca89:
Fix #776 
Fix #795 
Resolves ENG-2346
Resolves ENG-2299

This PR will fix a bug in `ssh_helper`'s `track-activity` that was not updating correctly the activity file in some situations,
also we standardize the `/tmp` in tmpfs as in many other linux distributions

This PR will also introduce a little fleet-server helper, in order to keep track of the fleet's connection activities and update the devpod's activity file accordingly, thus making it respect the inactivity timeout

### DIFF
--- a/cmd/agent/container/daemon.go
+++ b/cmd/agent/container/daemon.go
@@ -43,10 +43,12 @@ func (cmd *DaemonCmd) Run(_ *cobra.Command, _ []string) error {
 		return nil
 	}
 
-	err = os.WriteFile(agent.ContainerActivityFile, nil, 0777)
+	err = os.WriteFile(agent.ContainerActivityFile, nil, 0o777)
 	if err != nil {
 		return err
 	}
+
+	_ = os.Chmod(agent.ContainerActivityFile, 0o777)
 
 	// query the activity file
 	for {
--- a/cmd/agent/container/daemon.go
+++ b/cmd/agent/container/daemon.go
@@ -48,7 +48,10 @@ func (cmd *DaemonCmd) Run(_ *cobra.Command, _ []string) error {
 		return err
 	}
 
-	_ = os.Chmod(agent.ContainerActivityFile, 0o777)
+	err = os.Chmod(agent.ContainerActivityFile, 0o777)
+	if err != nil {
+		return err
+	}
 
 	// query the activity file
 	for {
--- a/cmd/helper/fleet_helper.go
+++ b/cmd/helper/fleet_helper.go
@@ -36,33 +36,38 @@ func NewFleetServerCmd(flags *flags.GlobalFlags) *cobra.Command {
 }
 
 // Run runs the command logic
-func (cmd *FleetServerCmd) Run(_ *cobra.Command, _ []string) error {
+func (c *FleetServerCmd) Run(cmd *cobra.Command, _ []string) error {
 	logFile := filepath.Join(os.Getenv("HOME"), ".cache/JetBrains/Fleet/log/fleet.log")
 	firstConnection := regexp.MustCompile(`.*Received authorization request.*`)
 	connStatus := regexp.MustCompile(`.*Notify.*`)
 
 	for {
-		time.Sleep(time.Second * 10)
+		select {
+		case <-time.After(time.Second * 10):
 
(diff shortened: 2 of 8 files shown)


cmd/agent/container/daemon.go
```diff
@@ -43,11 +43,13 @@ func (cmd *DaemonCmd) Run(_ *cobra.Command, _ []string) error {
 		return nil
 	}
 
-	err = os.WriteFile(agent.ContainerActivityFile, nil, 0777)
+	err = os.WriteFile(agent.ContainerActivityFile, nil, 0o777)
 	if err != nil {
 		return err
 	}
 
+	_ = os.Chmod(agent.ContainerActivityFile, 0o777)
```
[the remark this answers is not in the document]
89luca89: Yup, done :+1:

cmd/helper/fleet_helper.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// Run runs the command logic
+func (cmd *FleetServerCmd) Run(_ *cobra.Command, _ []string) error {
+	logFile := filepath.Join(os.Getenv("HOME"), ".cache/JetBrains/Fleet/log/fleet.log")
+	firstConnection := regexp.MustCompile(`.*Received authorization request.*`)
+	connStatus := regexp.MustCompile(`.*Notify.*`)
+
+	for {
+		time.Sleep(time.Second * 10)
```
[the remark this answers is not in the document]
89luca89: Yep you're right, much better this other way :+1:

pkg/driver/docker/docker.go
```diff
@@ -251,10 +251,20 @@ func (d *dockerDriver) RunDockerDevContainer(
 	}
 
 	// mounts
+	tmpfsMount := true
 	for _, mount := range options.Mounts {
+		// skip tmpfs mount on /tmp if devcontainer specifies another mount option
+		if strings.Contains(mount.String(), "/tmp") {
```
pascalbreuninger, line 257: maybe need a regex here? So we don't hit `/foo/tmp` cases

cmd/helper/fleet_helper.go
```diff
@@ -0,0 +1,73 @@
...
+
+// FleetServerCmd holds the fleet server cmd flags
+type FleetServerCmd struct {
+	*flags.GlobalFlags
+
+	WorkspaceID string
+}
+
+// NewFleetServerCmd creates a new fleet command
+func NewFleetServerCmd(flags *flags.GlobalFlags) *cobra.Command {
```
pascalbreuninger, line 23: I'm not sure if this should be a separate command tbh. Is there something speaking against moving it into the regular daemon?

pascalbreuninger (changes requested):

pkg/driver/docker/docker.go
89luca89: makes sense yes :+1:  on it

pascalbreuninger (approved):

PR #898, opened 2024-02-12
feat: start workspace in a subpath

89luca89:
Initial draft to support specifying a subpath where to start a workspace

When specified the subpath will be used as:

- default workdir via ssh
- default workspace for IDEs
- default path where to search for devcontainer.json

Full project will anyway be cloned and mounted into the container

Useful for monorepos

Resolves ENG-1976

I'm still not sure if this is the best way to do it, but it works for now, want to discuss here the implementation in the GUI, and if there are missing pieces

EDIT: to change subfolder, you use `--recreate` also, so it can switch also container

### DIFF
--- a/e2e/tests/ssh/ssh.go
+++ b/e2e/tests/ssh/ssh.go
@@ -50,7 +50,7 @@ var _ = DevPodDescribe("devpod ssh test suite", func() {
 			framework.ExpectNoError(err)
 		})
 
-		ginkgo.It("should start a new workspace with a docker provider (default) and forward gpg agent into it", func() {
+		ginkgo.FIt("should start a new workspace with a docker provider (default) and forward gpg agent into it", func() {
 			// skip windows for now
 			if runtime.GOOS == "windows" {
 				return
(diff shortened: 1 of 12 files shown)


pkg/provider/workspace.go
```diff
@@ -174,6 +174,8 @@ type CLIOptions struct {
 	DisableDaemon        bool     `json:"disableDaemon,omitempty"`
 	DaemonInterval       string   `json:"daemonInterval,omitempty"`
 
+	WorkspaceSubPath string `json:"workspacesubpath,omitempty"`
```
ThomasK33, line 177:
```suggestion
	WorkspaceSubPath string `json:"workspaceSubPath,omitempty"`
```

ThomasK33 (approved):
Lgtm

ThomasK33:
This looks good if we want to fork the entire repo.

For larger monorepos we might be interested in implementing sparse checkouts.

pkg/provider/workspace.go
89luca89: Thanks, fixed :+1:

89luca89:
> This looks good if we want to fork the entire repo.
> 
> For larger monorepos we might be interested in implementing sparse checkouts.

Yea was thinking about this but maybe as a separate option?

Like `--sparse-checkout SUBDIR` which will both do the sparse checkout and the subpath option 

This way it could be useful to have a full-checkout+subdir (maybe to access the rest of the project?), or a sparsecheckout that will automatically do also the subdir

ThomasK33:
Yeah, adding sparse checkout later makes more sense. You're right.

> Like --sparse-checkout SUBDIR which will both do the sparse checkout and the subpath option

I probably wouldn't do that. Assuming one might want to check out multiple folders, args should accept a comma-delimited list or paths to checkout, and then one can use the flag from this PR to change the working directory.

pascalbreuninger (approved):

ThomasK33 (approved):

pascalbreuninger (approved):

e2e/tests/ssh/ssh.go
```diff
@@ -80,6 +80,8 @@ var _ = DevPodDescribe("devpod ssh test suite", func() {
 			err = f.DevPodUp(devpodUpCtx, tempDir, "--gpg-agent-forwarding")
 			framework.ExpectNoError(err)
 
+			time.Sleep(time.Second * 10)
```
ThomasK33, line 83: Out of curiosity, why are we sleeping here?

89luca89: It was a test, I was suspecting some sort of racing/unready-ness because it was failing a bit randomly

I suspect at this point, it's a matter of the gpg agent not being "ready" for the forwarding, leading to some EOF errors during the forwarding

Will do a retry logic in place of this

pascalbreuninger (approved):

PR #905, opened 2024-02-16
feat: inject DEVPOD and REMOTE_CONTAINERS env variables

### LINKED ISSUE #891: Environment Variable to Detect Devcontainer
Hello! I wanted to make a suggestion that DevPod inject an environment variable that would allow scripts to detect they are running inside a devcontainer. VS Code did this [here](https://github.com/microsoft/vscode-remote-release/issues/3517) via setting REMOTE_CONTAINERS=true, doing the same would be convenient, maybe even adding DEVPOD=true to boot.

Thanks for a great product!

No

Inject the following env vars:
REMOTE_CONTAINERS=true
DEVPOD=true

None I'm aware of.

**Additional context**

89luca89:
Inject the following env variables in the container:

REMOTE_CONTAINERS=true
DEVPOD=true

This will ensure that scripts and software running inside can detect that we are in a remote, devpod, container.

Fix #891 
Resolves ENG-2793

### DIFF
--- a/pkg/driver/docker/docker.go
+++ b/pkg/driver/docker/docker.go
@@ -199,6 +199,11 @@ func (d *dockerDriver) RunDockerDevContainer(
 		"--sig-proxy=false",
 	}
 
+	// add environment variables that signals that we are in a remote container
+	// (vscode compatibility) and specifically that we are using devpod.
+	args = append(args, "--env", "DEVPOD=true")
+	args = append(args, "--env", "REMOTE_CONTAINERS=true")
+
 	// add ports
 	for _, appPort := range parsedConfig.AppPort {
 		intPort, err := strconv.Atoi(appPort)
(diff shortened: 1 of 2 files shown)


pkg/driver/docker/docker.go
```diff
@@ -199,6 +199,11 @@ func (d *dockerDriver) RunDockerDevContainer(
 		"--sig-proxy=false",
 	}
 
+	// add environment variables that signals that we are in a remote container
+	// (vscode compatibility) and specifically that we are using devpod.
+	args = append(args, "--env", "DEVPOD=true")
+	args = append(args, "--env", "REMOTE_CONTAINERS=true")
```
pascalbreuninger, line 205: If you set this in `driver.RunOptions.Env` it'll also work for custom drivers

pascalbreuninger (changes requested):

pkg/driver/docker/docker.go
89luca89: Done :+1:

pascalbreuninger (approved):

PR #908, opened 2024-02-18
otherPortsAttributes should be the attributes directly

justinsb:
portAttributes are keyed by port, otherPortsAttributes are the
fallback for unnamed ports.


pascalbreuninger (approved):
LGTM, thanks for the fix!

89luca89 (approved):

PR #917, opened 2024-02-20
fix: fix recreate logic if workspace is local folder

mukerjee:
If a workspace is created from an actual local folder (as opposed to downloading a git repo) _and_ we're using a local provider, then `--recreate` should not try to download that folder again.

Without this fix I was seeing an EOF issue on recreate (likely files were trying to be copied to / from the same destination leading to file corruption). For example, my `.git/index` file was being corrupted by this issue, leading to git commands causing a segfault.

### DIFF
--- a/cmd/agent/workspace/up.go
+++ b/cmd/agent/workspace/up.go
@@ -197,6 +197,12 @@ func prepareWorkspace(ctx context.Context, workspaceInfo *provider2.AgentWorkspa
 
 		return nil
 	} else if workspaceInfo.Workspace.Source.LocalFolder != "" {
+		// if we're not sending this to a remote machine, we're already done
+		if workspaceInfo.ContentFolder == workspaceInfo.Workspace.Source.LocalFolder {
+			log.Debugf("Local folder with local provider; skip downloading")
+			return nil
+		}
+
 		log.Debugf("Download Local Folder")
 		return DownloadLocalFolder(ctx, workspaceInfo.ContentFolder, client, log)
 	} else if workspaceInfo.Workspace.Source.Image != "" {
--- a/e2e/framework/command.go
+++ b/e2e/framework/command.go
@@ -100,6 +100,17 @@ func (f *Framework) DevPodUp(ctx context.Context, additionalArgs ...string) erro
 	return nil
 }
 
+func (f *Framework) DevPodUpRecreate(ctx context.Context, additionalArgs ...string) error {
+	upArgs := []string{"up", "--recreate", "--debug", "--ide", "none"}
+	upArgs = append(upArgs, additionalArgs...)
+
+	_, _, err := f.ExecCommandCapture(ctx, upArgs)
+	if err != nil {
+		return fmt.Errorf("devpod up --recreate failed: %s", err.Error())
+	}
+	return nil
+}
+
 func (f *Framework) DevPodSSH(ctx context.Context, workspace string, command string) (string, error) {
 	out, err := f.ExecCommandOutput(ctx, []string{"ssh", workspace, "--command", command})
 	if err != nil {
--- a/e2e/tests/up/up.go
+++ b/e2e/tests/up/up.go
@@ -277,6 +277,36 @@ var _ = DevPodDescribe("devpod up test suite", func() {
 			framework.ExpectNoError(err)
 		})
 
+		ginkgo.It("recreate a local workspace", func() {
+			const providerName = "test-docker"
+			ctx := context.Background()
+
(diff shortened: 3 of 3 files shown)


pascalbreuninger:
I do see the problem but this keeps the workspace content stale then. What do you think of removing the original folder and downloading a fresh copy for local folders?

89luca89:
What I think should be the correct behaviour, is that when `--recreate` we wipe the remote directory, and fresh copy again from local, as "local always wins" strategy

This way we don't have mixed/corrupt `.git`, and we don't have stale content

89luca89 (changes requested):
please check the comment

mukerjee:
I'm not quite sure I understand. Let me explain my usecase:
- I'm playing around with devpod to see if devcontainers make sense for my team's project.
- Our project repo didn't have a devcontainer.json, so I'm writing a devcontainer.json as I'm testing devpod.
- I want to rapidly iterate on my devcontainer.json, so I've copied my project's git repo to my local mac laptop and just bring up the container with `devpod up .` (with the local docker provider)
- After I make changes to devcontainer.json or my Dockerfile, I want to recreate the devpod, so i do `devpod up . --recreate`.

In this scenario there isn't a remote URL to git clone a repo from (as i'm using a local folder), nor is there a remote machine to copy files to (as i'm using the local docker provider).

This usecase was supported before https://github.com/loft-sh/devpod/commit/b749ce5e745cfbc64c0d89960d53e361cae5df0c went in, but now seems to cause git corruption.

My current change says something like "If using a local folder, don't force a copy on recreate", but perhaps it should say something like "if using a local folder && not using a remote machine, don't force a copy on recreate". Would that make more sense?

Thanks for the quick responses on this! really appreciate it.

89luca89:
mmh I find it strange that using a local provider (like docker) triggers the folder copy

with local providers it should simply mount the directory as a volume as you can see here in this example:
Screenshot from 2024-02-21 09-14-32

Copy of the workspace/repo dir is done for remote providers (like kubernetes, machine or ssh providers)
In that case, what I would do for `--recreate` is to have a "local dir always wins" policy, so remote dir is deleted and copied again

Your fix would work, but at one point the remote workspace would be "stale", so it would risk to be older than the local one, that's why I'm proposing this change

mukerjee:
Thanks for trying this out. Just to verify, if you do to the same thing you did above, but add `--recreate`, do you see the issue I mentioned (i.e., your local folder is accidentally "uploaded to server" even though there is no remote server + you get git corruption)? i.e., try:
```
./devpod up --ide=none --provider=docker ~/Projects/Personal/distrobox --recreate
```

"local dir always wins" makes sense -- can you take a look at my updated change? I'm not sure if `workspaceInfo.Workspace.Machine.ID` is the right variable to check as I'm only using a local provider. That said, this still fixes the issue for my workflow.

89luca89:
@mukerjee I think you're right, I can reproduce the problem with local `--recreate`

I think these are 2 separate issues

1 - local provider should skip the folder upload on recreate
2 - remote providers should delete remote folder and redo the upload

--- 

For 1, I'd check if `workspaceInfo.ContentFolder` is equal to `workspaceInfo.Workspace.Source.LocalFolder`, that usually means we're using a simple bind mount, and we need to skip

For 2, I'd need to investigate a bit

89luca89:
Yep this works, want to take a shot at problem number 2? :grin:

mukerjee:
> Yep this works, want to take a shot at problem number 2? 😁

Great-- thanks for taking a look!

I can't do 2 yet unfortunately. I don't have a remote platform to test with yet. :(

This PR should solve my workflow problem at least, so if we could merge it and cut a new alpha release, that'd be appreciated!

89luca89:
@mukerjee that's ok we can open a new issue for case-2

Last request, if you can add an e2e test for this PR?
It will just need to do `up` dn `up --recreate`, you can find many examples in `e2e/tests/up/up.go` to copy from :innocent:

mukerjee:
> @mukerjee that's ok we can open a new issue for case-2
> 
> Last request, if you can add an e2e test for this PR? It will just need to do `up` dn `up --recreate`, you can find many examples in `e2e/tests/up/up.go` to copy from 😇

Sounds good -- I added two e2e tests. I was having trouble running them locally, so I'm hoping to just see them pass on CI :)
Let me know if you want any changes to the tests themselves. happy to adjust.

89luca89 (approved):

89luca89:
Thanks @mukerjee great job! :heart:

PR #919, opened 2024-02-21
fix: Windows Runner cleanup

89luca89:
Fix Windows Runner piling up images with `devpod` and `none` tags

Screenshot from 2024-02-21 10-26-37


89luca89:
After new cleanup fix

image

PR #923, opened 2024-02-23
feat: add build cli options for buildx

mukerjee:
cli build options were recently added to the devcontainer spec (see https://github.com/devcontainers/spec/commit/d424cc157e9a110f3bf67d311b46c7306d5a465d ). let's add them here as well.

This can be useful for example to add a `--secret ...` to a `docker build`.

It's a little unclear currently how to add these to `pkg/devcontainer/buildkit/buildkit.go`. For now, I left a TODO. This would be no worse than the current behavior (as currently `build.options` is just ignored).

### DIFF
--- a/pkg/devcontainer/build/options.go
+++ b/pkg/devcontainer/build/options.go
@@ -4,6 +4,8 @@ type BuildOptions struct {
 	BuildArgs map[string]string
 	Labels    map[string]string
 
+	CliOpts []string
+
 	Images    []string
 	CacheFrom []string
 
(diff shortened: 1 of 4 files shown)


pascalbreuninger:
Fantastic, thank you 🙌 let me check if I could quickly sort out the BuildKit options before we merge this

pascalbreuninger:
Update: skimmed over the buildx codebase to see how they resolve the CLI flags to buildkit options but there's too much of a difference for an easy conversion. Let's keep the TODO and see if it ever becomes an issue

pascalbreuninger (approved):

mukerjee:
Thanks @pascalbreuninger ! I've fixed the linter complaints, so once CI runs again, this should be good to merge.

pascalbreuninger:
Great, thanks! Will merge once Pipelines are done

PR #934, opened 2024-02-28
Feat/recreate for drivers

### LINKED ISSUE #166: Kubernetes driver rebuild support
ul

89luca89:
With this PR we add `--recreate` support for custom drivers

Fix #166 
Resolves ENG-1367

### DIFF
--- a/pkg/devcontainer/run.go
+++ b/pkg/devcontainer/run.go
@@ -92,8 +92,7 @@ func (r *runner) Up(ctx context.Context, options UpOptions) (*config.Result, err
 	// download workspace source before recreating container
 	_, isDockerDriver := r.Driver.(driver.DockerDriver)
 	if options.Recreate && !isDockerDriver {
-		// TODO: implement this
-		return nil, fmt.Errorf("rebuilding the workspace is currently not supported for non-docker drivers")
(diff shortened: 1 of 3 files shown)


pkg/devcontainer/run.go
```diff
@@ -316,6 +315,17 @@ func runInitializeCommand(
 	return nil
 }
 
+func (r *runner) recreateCustomDriver(ctx context.Context, options UpOptions) (*config.Result, error) {
+	err := r.Driver.StopDevContainer(ctx, r.ID)
+	if err != nil {
+		return nil, errors.Wrap(err, "rebuilding the workspace is currently not supported for this driver")
```
pascalbreuninger, line 321: not sure the error message makes sense now 🤔

89luca89: Welp in case we will have other custom drivers where recreate doesn't work :shrug:

pascalbreuninger: but it could also fail for unrelated reasons, because all custom drivers can be stopped

pascalbreuninger: sounds good

89luca89: done :+1:

PR #971, opened 2024-03-22
fix(test): fix agent url on non-linux runners

89luca89:
add way for non-linux runners to download a freshly compiled devpod binary as agent

Resolves ENG-3057

### DIFF
--- a/.github/workflows/e2e-win-full-tests.yaml
+++ b/.github/workflows/e2e-win-full-tests.yaml
@@ -20,15 +20,13 @@ jobs:
       max-parallel: 1
       matrix:
         label:
-          [
-            "build",
-            "ide",
-            "ssh",
-            "up-docker",
-            "up-docker-build",
-            "up-docker-compose",
-            "up-docker-wsl",
-          ]
+          - "build"
+          - "ide"
+          - "ssh"
+          - "up-docker"
+          - "up-docker-build"
+          - "up-docker-compose"
+          - "up-docker-wsl"
 
     steps:
     - name: Git set line ending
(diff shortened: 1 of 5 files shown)


.github/workflows/e2e-win-full-tests.yaml
```diff
@@ -19,8 +19,16 @@ jobs:
...
+        label:
+          [
+            "build",
+            "ide",
+            "ssh",
+            "up-docker",
+            "up-docker-build",
+            "up-docker-compose",
+            "up-docker-wsl",
+          ]
```
ThomasK33, line 31: Would it make sense to change this to a YAML-style list?

89luca89: Makes sense, done :+1:

e2e/framework/server_utils.go
```diff
@@ -0,0 +1,108 @@
...
+	// Create a file server handler for the specified directory
+	fileServer := http.FileServer(http.Dir(dir))
+
+	// Register the file server handler to serve files under the /files route
+	http.Handle("/files/", http.StripPrefix("/files", fileServer))
+
+	randomPort, err := findOpenPort(0)
+	if err != nil {
+		log.Fatal(err)
+	}
```
ThomasK33, line 36: Is there a hard constraint on having a port between 10k and 40k?

Otherwise, we could replace this port-finding process with a call to `listener, err := net.Listen("tcp", fmt.Sprintf("%v:0", ip))`.
That should return a listener that `http.Serve` can consume, and we can grab the port by using `listener.Addr().(*net.TCPAddr).Port`

89luca89: No constraints, indeed your suggested solution simplifies things a lot, thanks for pointing it out! :smile:

89luca89: Done :+1:

ThomasK33 (approved):

PR #978, opened 2024-03-27
feat: add support for devcontainer up --workspace-env-file

### LINKED ISSUE #827: Add support for devcontainer up --secrets-file

yes

The workspace configuration should allow configuring a secrets file. This should be passed to the `devcontainer up` invocation, regardless of provider.

None.

**Additional context** 

This would enable us to more easily control secrets (as opposed to https://devpod.sh/docs/developing-in-workspaces/environment-variables-in-devcontainer-json). We need secrets at runtime and during prebuilds, for accessing external systems like Codeartifact.

eduardodbr:
Fixes #827

Each file from the list is parsed for key value pairs in the format `key=value` and merged with the remaining `--workspace-env` variables.

### DIFF
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"bufio"
 	"bytes"
 	"context"
 	"fmt"
@@ -12,7 +11,6 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
-	"unicode/utf8"
 
 	"github.com/loft-sh/devpod/cmd/flags"
 	"github.com/loft-sh/devpod/pkg/agent"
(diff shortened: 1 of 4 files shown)


cmd/up.go
```diff
@@ -740,6 +747,52 @@ func mergeDevPodUpOptions(baseOptions *provider2.CLIOptions) error {
...
+			return err
+		}
+		variables = append(variables, envFromFile...)
+	}
+	baseOptions.WorkspaceEnv = append(baseOptions.WorkspaceEnv, variables...)
+
+	return nil
+}
+
+func parseKeyValueFile(filename string) ([]string, error) {
```
eduardodbr, line 764: First time contributing so I'm not very familiar with the codebase, let me know if there's a generic package to put this bit in

89luca89: Hi @eduardodbr you could put this parse function in `pkg/devcontainer/config/` package :+1:

e2e/tests/up/up.go
```diff
@@ -82,6 +82,41 @@ var _ = DevPodDescribe("devpod up test suite", func() {
...
+			validData := []byte("TEST_VAR=" + value)
+			workspaceEnvFileValid := filepath.Join(tmpDir, ".valid")
+			err = os.WriteFile(
+				workspaceEnvFileValid,
+				validData, 0o644)
+			framework.ExpectNoError(err)
+			defer os.Remove(workspaceEnvFileValid)
+
+			// set env var
+			err = f.DevPodUp(ctx, name, "--workspace-env-file", workspaceEnvFileValid)
```
89luca89, line 113: can we add a test case for multiple env files?

89luca89 (changes requested):

89luca89:
Thanks a lot @eduardodbr for the PR!

Just a couple of adjustments :smile:  great work, thanks!

cmd/up.go
eduardodbr: done, thanks

e2e/tests/up/up.go
eduardodbr: done

eduardodbr:
> Thanks a lot @eduardodbr for the PR!
> 
> Just a couple of adjustments 😄 great work, thanks!

Done the suggested changes, thanks. Also changed `workspace-env-file` to be StringSliceVar instead of StringArrayVar, this way it already parses the arguments as comma-separated list of filenames

89luca89 (approved):

89luca89:
Thanks a lot @eduardodbr LGTM, running the test suite, and if all is green will merge :+1:

89luca89:
Thanks a lot!

PR #1001, opened 2024-04-12
Fix/SSH Conf: Generated ProxyCommand line is missing quotes around path

landoncolburn:
Fails to connect to DevPod on Windows over SSH since path for ProxyCommand in outputted `.ssh/config` is grabbed via `os.Executable()`, the default install location on Windows (`C:\Program Files`) has a space resulting in line:

```
  ProxyCommand C:\Program Files\.... ssh --stdio --context default --user devpod workspace-1"
```

Fixed by wrapping the path in escaped quotations

### DIFF
--- a/pkg/ssh/config.go
+++ b/pkg/ssh/config.go
@@ -67,11 +67,11 @@ func addHost(path, host, user, context, workspace, workdir, command string, gpga
 	newLines = append(newLines, "  UserKnownHostsFile /dev/null")
 	newLines = append(newLines, "  HostKeyAlgorithms rsa-sha2-256,rsa-sha2-512,ssh-rsa")
 	if command != "" {
-		newLines = append(newLines, fmt.Sprintf("  ProxyCommand %s", command))
+		newLines = append(newLines, fmt.Sprintf("  ProxyCommand \"%s\"", command))
 	} else if gpgagent {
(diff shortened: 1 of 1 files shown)


shanman190:
This one would be good for Windows users that don't happen to be using the OpenSSH build that ships internal to Windows 10/11. The internal Microsoft build hides this nuance, but other SSH builds such as the one that ships with Git Bash do not perform any tricks related to the handling of spaces in the path and error out instead.

landoncolburn:
> This one would be good for Windows users that don't happen to be using the OpenSSH build that ships internal to Windows 10/11. The internal Microsoft build hides this nuance, but other SSH builds such as the one that ships with Git Bash do not perform any tricks related to the handling of spaces in the path and error out instead.

Ah that clears some stuff up! I didn't realize that the OpenSSH build handles path resolution differently than Git Bash one, but that makes sense why I was only experiencing this issue on some machines. All versions that I tested support the quoted path though, so this might be a good option even if just for legacy purposes.

shanman190:
Yep! The quoted variations are supported universally regardless of OS, so in my opinion this is still a good fix.

pascalbreuninger (approved):
LGTM, thanks for the PR!

Also didn't know the default windows ssh executable handles whitespace in paths differently

PR #1002, opened 2024-04-13
chore: update IDE versions

shanman190:


pascalbreuninger (approved):
LGTM, thanks for the PR

shanman190:
@pascalbreuninger, something else that I was wondering too with respect to this. Would you all be open the the virtual urls like was done for fleet?

https://github.com/loft-sh/devpod/blob/main/pkg%2Fide%2Ffleet%2Ffleet.go#L34-L45

This would enable always ensuring that DevPod is downloading the latest available JetBrains UI when it is provisioning a workspace.

pascalbreuninger:
@shanman190 of course, it just hasn't been a big enough problem for us yet. If you're willing to open a PR we'd be happy to merge that

RobinQu:
When will this go into next release then?

pascalbreuninger:
@RobinQu I'd wait 1-2 weeks because the 2024.1 updates are pretty new still

shanman190:
@pascalbreuninger, I've created https://github.com/loft-sh/devpod/pull/1006 like we discussed earlier.

nojaf:
Is this available in any alpha?

pascalbreuninger:
@nojaf yes, should be in the latest one

shanman190:
@nojaf, the very latest alpha should download the latest version by default. Then you can set the version explicitly as well to choose an alternative version.

PR #1006, opened 2024-04-15
Download latest version of Jetbrains tools by default

shanman190:
Additionally, retain override hook from `~/.devpod/config.yaml`.

Product codes can be found here:
https://youtrack.jetbrains.com/issue/MP-2363/Can-you-provide-me-the-list-of-jetbrains-product-codes

### DIFF
--- a/pkg/ide/jetbrains/clion.go
+++ b/pkg/ide/jetbrains/clion.go
@@ -7,6 +7,7 @@ import (
 )
 
 const (
+	CLionProductCode           = "CL"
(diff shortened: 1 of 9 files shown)


shanman190:
Another aspect here would be potentially exposing a way for the user to explicitly specify the `VERSION` config value in the UI rather than needing to edit their `~/.devpod/config.yaml` file.

pascalbreuninger (approved):
@shanman190 thanks for the PR, looks like a good solution to always keep it up to date 👍 
I completely forgot we added this because it's been a while but you can always set the IDE version like so:
`devpod ide set-options intellij -o VERSION=2024.1`

Issue #875, opened 2024-01-23
gosshagent.ForwardToRemote fails with ssh agent from git for windows


adastraperangusta:
Hello,

I am using DevPod v0.5.0-alpha.1 on windows (build 1607) and I am using ssh client from git.

**What happened?**  

I'm getting the following error every time i try to set up a workspace (with any provider like ssh or docker)

```log
[14:41:38] info Workspace src-1 already exists
[14:41:38] debug Acquire workspace lock...
[14:41:38] debug Acquired workspace lock...
[14:41:38] info Creating devcontainer...
[14:41:38] debug Inject and run command: '/tmp/devpod/agent' helper ssh-server --stdio --debug
[14:41:38] debug execute inject script
[14:41:38] debug download agent from https://github.com/loft-sh/devpod/releases/download/v0.5.0-alpha.1
[14:41:38] debug Attempting to create SSH client
[14:41:38] debug Run command provider command: ${SSH_PROVIDER} command
[14:41:46] debug Received line after pong: done
[14:41:46] debug SSH client created
[14:41:46] debug SSH session created
[14:41:47] info Execute SSH server command: bash -c '/tmp/devpod/agent' agent workspace up --workspace-info '....' --debug
[14:41:47] info No root required, because neither docker nor agent daemon needs to be installed
[14:41:47] info Use /home/user/.devpod/agent/contexts/default/workspaces/src-1 as workspace dir
[14:41:47] debug Created logger
[14:41:47] debug Received ping from agent
[14:41:47] debug Credentials server started on port 16886...
[14:41:48] debug Incoming client connection at /
[14:41:48] debug Credentials server started...
[14:41:48] debug Wrote docker credentials helper to /home/user/.devpod/agent/contexts/default/workspaces/src-1/.cache/hmofachtfmjr/docker-credential-devpod
[14:41:48] debug Workspace Folder already exists /home/user/.devpod/agent/contexts/default/workspaces/src-1/content
[14:41:48] debug Workspace exists, skip downloading
[14:41:48] debug Using docker command 'docker'
[14:41:48] info find dev container: docker ps: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
[14:41:48] info exit status 1
[14:41:48] info devcontainer up
[14:41:48] info github.com/loft-sh/devpod/cmd/agent/workspace.(*UpCmd).Run
[...]

PR #1008, opened 2024-04-16
Enable SSH agent support on Windows

shanman190:
The purpose of this PR is to enable SSH agent support universally throughout DevPod. As mentioned in https://github.com/loft-sh/devpod/issues/875 the upstream golang.org/x/crypto/ssh/agent library doesn't support named pipes and only supports unix sockets. This results in any usage requiring SSH agent forwarding -- outside of using the native SSH client directly -- not being enabled, even if requested.

While it would be ideal for the upstream library to incorporate named pipes support, it's been frozen in place since early 2021 awaiting the inclusion of golang.org/x/sys/windows/namedpipe (via golang/go#49650). However, the aforementioned issue has been slow in getting adopted due to Microsoft copyright review due to some included components on that PR in particular.

As a result of all of this, this PR introduces a thin shim that coordinates SSH Agent forwarding between the upstream golang.org/x/crypto/ssh library for unix systems and a mirror implementation of the golang SSH agent support, but using named pipes.


shanman190:
For the why is this important, presently without SSH agent forwarding working on Windows this means that DevPod can't seamlessly provide transitive identity into the machine or workspace using it's native implementation. This results in being unable to perform SSH-based Git commands against repositories and being forced to fallback to either:

* Workspace Init: a local clone, then DevPod copies the local content to the container (This typically results in EOL issues as Windows clones CRLF by default)
* Ongoing Git operations: the user is either forced to setup an existing or new SSH key within the container to then perform SSH operations OR use an alternative authentication method.

shanman190:
Also there were two more places where `SSH_AUTH_SOCK` appears, but it wasn't super clear where those invocations originate from, so for the purposes of this change I left those unmodified.

pkg/ssh/agent/agent_windows.go
```diff
@@ -0,0 +1,80 @@
...
+	sshAuthSocket := os.Getenv("SSH_AUTH_SOCK")
+	if sshAuthSocket != "" {
+		return sshAuthSocket
+	}
+	if _, err := os.Stat(defaultNamedPipe); err == nil {
+		return defaultNamedPipe
+	}
+
+	return ""
+}
```
shanman190, line 29: I'm planning to test this just a little bit more tomorrow, but presently I know this works with the OpenSSH agent that ships with Windows 10/11 natively. I'm pretty sure that the Git for Windows variation utilizes a unix socket which would mean a slight adjustment to the agent forwarding to ensure the correct socket type is used for the connection.

pascalbreuninger:
@shanman190 thanks for the PR, we'll look into it soon

pkg/ssh/agent/agent_windows.go
shanman190: I've confirmed that Git for Windows is in fact using an `AF_UNIX` socket, so I'll update the forward method just a little bit to handle that.

shanman190: After some more digging after being puzzled why I couldn't connect using the AF_UNIX socket type as directly utilized by the upstream `golang.org/x/crypto/SSH/agent`, I've come to find out that MSYS2 which Got for Windows is based on uses an emulated socket rather than a native AF_UNIX socket as I was led to believe. 

Now with this said, there are a myriad of authenticators that do create valid AF_UNIX sockets for Windows and I found a PR for Win32-OpenSSH to add native AF_UNIX sockets to the build that comes packaged with Windows. Unfortunately, [that PR](https://github.com/PowerShell/openssh-portable/pull/674) is not yet merged.

What I think I'm going to do here is to prefer a set SSH_AUTH_SOCK, this way various correct implementations work, then fallback to the named pipe implementation, then finally return an error. When Windows finally natively supports SSH_AUTH_SOCK definitions, then the code will natively migrate to that updated path this way.

I'll submit this small change tomorrow, but then this should be good to go.

pascalbreuninger:
@shanman190 first up, thanks a lot for your work on this issue and the detailed explanation!
Both @89luca89 and me aren't Windows users and aren't really familiar with the environment, we appreciate all the help we can get there. 

> Also there were two more places where SSH_AUTH_SOCK appears, but it wasn't super clear where those invocations originate from, so for the purposes of this change I left those unmodified.

I took a look and they are inside of the container, should be fine 👍

shanman190:
@pascalbreuninger, you're welcome! Thanks to both you and @89luca89 taking a look as well. I think that I've got everything here working in a way that should work out very nicely for us all. 😄

pascalbreuninger (approved):
LGTM 🙌

shanman190:
@pascalbreuninger, so everything here looks great on the latest alpha. One quick question though that I have with ssh clones still failing is that I'm seeing it clone with the Git credential helper enabled even though my AWS provider has it explicitly disabled. So far I haven't been able to come up with any set of combinations that will actually disable it. Do you have any ideas?

pascalbreuninger:
@shanman190 great 👍 

Maybe try the context option `devpod context set-options -o SSH_INJECT_GIT_CREDENTIALS=false`
This is a problem of specificity (user option > provider option). I'm curious, would you expect this to be the other way round and if so, why? 
[ResolveAgentConfig](https://github.com/loft-sh/devpod/blob/c3f49b73015f2a9d2b182a04b99578299a87609c/pkg/options/resolve.go#L240)

shanman190:
@pascalbreuninger, so based on https://github.com/loft-sh/devpod/blob/c3f49b73015f2a9d2b182a04b99578299a87609c/pkg%2Foptions%2Fresolve.go#L216 it feels kind of backwards.

In my mind the precedence kinda feels like it should be:

* (lowest) provider defaults (SSH, AWS, Kubernetes, etc)
* user preferences `~/.devpod/config.yaml`
* machine, if applicable
* (highest) workspace

I think that's just an illustration of the same thing that you said though, but please correct me if I'm wrong.

To me though in the order that I've laid out it's furthest away from the workspace to the closest. The closer to the workspace the setting is the more prevalent I would think that option would become.

For my specific setup, I'm setting `INJECT_GIT_CREDENTIALS` and `INJECT_DOCKER_CREDENTIALS`, 

* Provider - AWS: git=true, docker=true
* User: git=false, docker=false
* Machine: git=false, docker=false (looks like it cloned the settings from the user config when the machine was first created)
* Workspace: none configured

NOTE: `agent.injectGitCredentials` as a setting only appears in the AWS provider.yaml, but since it's using the property placeholder I would think that it's resolution would be delayed as late as possible.

So I would envision the machine level taking precedence, even though I didn't explicitly set those (the machine was created as a side effect of creating the workspace in the DevPod UI), resulting in those settings being false when creating the workspace.

shanman190:
Alright, I've discovered that the helper isn't my problem. For whatever reason, I was missing that DevPod `sudo`'s to `root` where it then clones. `sudo` doesn't take the environment variables by default, but it can easily be told to do so. So I'll have another fix incoming that should finalize this feature and actually make ssh clones work everywhere.

PR #1027, opened 2024-04-25
Preserve environment variables through sudo session

shanman190:
When performing ssh clones on a machine where the Docker socket requires administrative access to utilize passing over the environment variables into the `sudo` session is required. 

It should also be noted, that Windows ships with OpenSSH 8.1 by default, so if Windows users are having issues with SSH Agent support they may need to upgrade their OpenSSH distribution to be inline or newer than the target machine's environment.

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -288,7 +288,7 @@ func rerunAsRoot(workspaceInfo *provider2.AgentWorkspaceInfo, log log.Logger) (b
 	}
 
 	// call ourself
-	args := []string{binary}
+	args := []string{"--preserve-env", binary}
 	args = append(args, os.Args[1:]...)
 	log.Debugf("Rerun as root: %s", strings.Join(args, " "))
 	cmd := exec.Command("sudo", args...)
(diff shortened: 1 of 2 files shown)


pkg/agent/agent.go
```diff
@@ -288,7 +288,7 @@ func rerunAsRoot(workspaceInfo *provider2.AgentWorkspaceInfo, log log.Logger) (b
 	}
 
 	// call ourself
-	args := []string{binary}
+	args := []string{"--preserve-env", binary}
```
shanman190, line 291: This preserves all environment variables from the initial session into the `sudo` session. Technically, this could be reduced down to `--preserve-env=SSH_AUTH_SOCK` to limit only the SSH Agent socket being passed into the `sudo` session.

shanman190: Just to note, the preservation of the environment variables is necessary regardless of source OS. This is all about the target OS and how the Docker socket has been configured from an access standpoint. So any user using a target where Docker requires admin rights would have been impacted by this.

pascalbreuninger (approved):
LGTM, thanks!

PR #1028, opened 2024-04-26
fix: quoted some more paths to handle spaces

landoncolburn:
I had some more time to look through the codebase and see if I can find any more paths that need to be quoted for support on Windows. I found another dozen or so (at least from what I can tell). Further to #1001


pascalbreuninger:
@landoncolburn going to close this PR. Afaik most issues on windows regarding path spaces should be resolved by now. Let us know if you find some that need pathing, though

PR #1029, opened 2024-04-26
fix: close resp body

testwill:


pascalbreuninger (approved):
Whoopsie, thanks @testwill

PR #1033, opened 2024-04-27
fix: fix feature install order

aacebedo:
When I tried to install features by specifying the order in their manifest I noticed it was not working (with local feature). I saw that the ID with which the ID was compared was not the ID of the manifest but the name given in the devcontainer.json.

This PR fixes it.


aacebedo:
@pascalbreuninger can you have a look please?

pascalbreuninger:
@aacebedo I'm a bit concern about using `Config.ID` in a function shared by both the automatic detection and manual feature order. It might make sense to change `computeFeatureOrder` instead of `computeAutomaticFeatureOrder` 
 
Could you add an example please?

pascalbreuninger:
@aacebedo thanks for the example!
Not sure I follow completely, I'll need a bit of time to take a look at the spec again. In the meantime would using `overrideFeatureInstallOrder` work for your use case? Just to get you unblocked until we merge a potential fix

PR #1080, opened 2024-05-23
fix: trim v from semver

anbraten:
Fixing `11:48:19 info Invalid character(s) found in major number "v2"` error received from `semver` package as my docker compose version is `v2.2.3`

### DIFF
--- a/pkg/compose/helper.go
+++ b/pkg/compose/helper.go
@@ -143,7 +143,7 @@ func (h *ComposeHelper) Remove(ctx context.Context, projectName string, args []s
 }
 
 func (h *ComposeHelper) GetDefaultImage(projectName, serviceName string) (string, error) {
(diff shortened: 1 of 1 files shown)


89luca89:
Hi @anbraten 
thanks for the PR!

This never has given problems, could it be a problem with newer compose version (that introduced the v?)
Maybe we can do a fallback instead, in case of error, then trim the V and try again?

anbraten:
Hi @89luca89 

I think this might have been a bug in the docker-compose plugin. It works for the latest `2.27.0` version for me, but some previous had issues when testing. 

However I still think the fix is worth it as `strings.TrimPrefix` will only remove the `v` prefix if it exists otherwise it will just pass the string to `semver.Parse` which therefore will work in both cases.

pascalbreuninger (approved):

PR #1115, opened 2024-06-07
fix(build): add flatpak support

89luca89:
This PR aims to fix some issues when running Devpod into a flatpak

### DIFF
--- a/desktop/src/client/client.ts
+++ b/desktop/src/client/client.ts
@@ -232,17 +232,17 @@ class Client {
 
   public async isCLIInstalled(): Promise<Result<boolean>> {
     try {
-      const result = await new Command("run-path-devpod-cli", ["version"]).execute()
       // we're in a flatpak, we need to check in other paths.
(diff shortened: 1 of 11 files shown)


desktop/src/client/client.ts
```diff
@@ -225,9 +225,24 @@ class Client {
...
+    return await invoke("get_env", { name });
+  }
+
   public async isCLIInstalled(): Promise<Result<boolean>> {
     try {
       const result = await new Command("run-path-devpod-cli", ["version"]).execute()
+      // we're in a flatpak, we need to check in other paths.
+      const isflatpak = await this.getEnv("FLATPAK_ID");
+      const home_dir = await this.getEnv("HOME");
+      if (!isflatpak) {
```
pascalbreuninger, line 239: shouldn't this be `if (isflatpak)`? 🤔

89luca89: fixed :+1:

pascalbreuninger (approved):

PR #1120, opened 2024-06-12
Remove redundant error checks

furkansenharputlu:
This PR removes the repeating redundant error checks:

```
err := someFunc()
if err != nil {
   return err
} 

return nil
```

Instead, directly returning removes redundancy and makes code more readable like this:

```
return someFunc()
```


furkansenharputlu:
This is very disrespectful. There is a contribution. Even you don’t merge, you write a comment why you closed.

pascal-breuninger:
Hey @furkansenharputlu, you’re right, sorry!
While we value the time you spent on the pr, it adds more noise than value in my opinion

PR #1139, opened 2024-06-28
Support Wolfi-based dev containers

imjasonh:
[Wolfi](https://wolfi.dev) is a relatively new distro focused on minimalism and fast package updates, in the service of low-to-no-CVE images. Like Alpine, it uses `apk` as its package manager. Unlike Alpine, it's glibc-based, and doesn't use musl.

I tried to use devpod to run a Wolfi-based container, and got this error:

Screenshot 2024-06-28 at 9 28 39 AM

Decoding that error, it says:

```
fetch https://packages.cgr.dev/extras/x86_64/APKINDEX.tar.gz
fetch https://packages.wolfi.dev/os/x86_64/APKINDEX.tar.gz
[https://packages.cgr.dev/extras]
[https://packages.wolfi.dev/os]
OK: 68154 distinct packages available
ERROR: unable to select packages:
 gcompat (no such package):
   required by: world[gcompat]
```

Looking at the code, it looks like `alpine.go` expects the presence of `apk` to determine whether the image is Alpine, and if so, it installs necessary tools, including `gcompat`. Wolfi, being glibc-based, doesn't include `gcompat` (it's a glibc compatibility layer for musl), so this fails.

This PR makes the following changes to better support Wolfi-based images:

1. check for the presence of `ID=alpine` in `/etc/os-release` to determine if `gcompat` is needed.
2. rename `InstallAlpineRequirements` to `InstallAPKRequirements` and rephrase error messages, since it now handles any distro using `apk` as a package manager.
3. fix the error message encoding

Let me know if there's anywhere else I should be looking for Alpine assumptions, I'm happy to send more PRs after this one.


pascalbreuninger:
Hey @imjasonh, thanks a lot for the PR, It's a very welcome fix for Wolfi (I've never tried it personally)!
We should only have this one area where we do a one-off for alpine and it's only because VSCode Server can't install without these dependencies. Other IDEs should be fine.

pascalbreuninger (approved):

PR #1143, opened 2024-06-29
feat(desktop): add color mode switch setting

giovanni-d:
I've been using devpod for a few days now but I was getting blind while using the desktop app :) 
I'm introducing a new color mode switch setting since I've noticed a lot of the styling already supports dark/light values. 

chakraui is the library of choice for the ui but there is a [problem](https://github.com/chakra-ui/chakra-ui/issues?q=is%3Aissue+is%3Aopen+flash) with the color mode option that causes the app to flash the light mode and then the dark mode (if you have the dark mode set beforehand and you open the app, vice versa). I saw there was already an attempt to fix this issue by removing the local storage key from the index.html but I introduced a new [check](https://github.com/giovanni-d/devpod/blob/65ad28ad2b6a958cb25a7381e88fcb836c95431b/desktop/src/Theme/theme.ts#L94) that should allow the app to start using the correct color mode saved (in local storage) or fallbacks to a default value of 'light' as it was configured.

Unfortunately getting the initial value for the color mode from the [settingsStore ](https://github.com/loft-sh/devpod/blob/65ad28ad2b6a958cb25a7381e88fcb836c95431b/desktop/src/contexts/SettingsContext.tsx#L49) in time for chakra to read from extendTheme is difficult right now with the current setup. I opted for retrieving from local storage which seems to be working fine during app restarts. the value from settingsStore is still being used on the settings page for the switch buttons. Let me know what you think of this approach. 

Note: when I generated the bindings for the new setting in rust, `dotfiles_url` got created as `dotfilesUrl` instead of `dotfilesURL`. I can add a `#[serde(rename = "dotfilesURL")]` if you want but I'm leaving these changes as is since the name now looks better.

EDIT: moved color mode setting to experimental section as discussed. new preview below :) 

https://github.com/loft-sh/devpod/assets/7089154/81fc0a71-2c1a-4e82-995f-69cbf3bacefb


pascalbreuninger:
Hi @giovanni-d, thanks for the PR. I highly appreciate your contribution. 
The reason why we've removed dark mode again is because it'd require more effort with providers and provider images then we were able to invest at the time.

Would you be okay with keeping the changes on your fork until we've figure this out? 

Alternatively we could also think about making this an experimental setting to set expectations for users right away

giovanni-d:
@pascalbreuninger thanks for checking this out. I would be down to release this as an experimental setting. At least users can have a taste for what's to come in the future :) How would we proceed with this approach? I can add a description below the setting name with something like: `Experimental feature. Use it at your own risk` 

Also could you tell me more about this problem with the provider images. Is this because not every provider has an image  that matches the dark mode theme? 🤔

pascalbreuninger:
> How would we proceed with this approach?

You'd need to prefix the setting with `experimental` (checkout the fleet setting as an example) and then move the setting into the experimental section and we should be fine 👍  

> Also could you tell me more about this problem with the provider images. Is this because not every provider has an image that matches the dark mode theme? 🤔

yes exactly, same goes for IDEs

giovanni-d:
@pascalbreuninger nice, okay I think I did it correctly. please check the new changes and the new preview video in the description. 

> yes exactly, same goes for IDEs

ah I understand now. if you guys need help with this in the future, I would be happy to help out :)

pascalbreuninger (approved):
LGTM, thanks for the contribution!

PR #1151, opened 2024-07-10
Update install.mdx

phreed:
The `-o` option is ambiguous. 

Invoke-WebRequest: Parameter cannot be processed because the parameter name 'o' is ambiguous. Possible matches include: -OperationTimeoutSeconds -OutFile -OutVariable -OutBuffer.


pascalbreuninger:
@phreed thanks for the clarfication 🙇

Issue #1146, opened 2024-07-04
Better way to support commit signing via SSH for SSH Provider


lukas-mertens:
**Is your feature request related to a problem?**  
Currently it is possible to sign commits via gpg, but git allows signing via ssh as well. There should be support for doing that as well for providers like the SSH Provider.

**Which solution do you suggest?**  
Provide a way to configure an individual .gitconfig which will be placed in the container + a way to securely put ssh public keys in the container. Maybe I am missing something, but this is the way I found which isn't **that** user-friendly:
https://garden.lukmert.de/30-resources/31-how-to/31-03-devpod/signing-commits-when-using-devpod-with-ssh-provider/

**Which alternative solutions exist?**  
See my workaround:
https://garden.lukmert.de/30-resources/31-how-to/31-03-devpod/signing-commits-when-using-devpod-with-ssh-provider/

**Additional context**  
We discussed a bit on slack: https://loft-sh.slack.com/archives/C056ZDZPJ4W/p1720003846623999

pascalbreuninger:
Hi @lukas-mertens, thanks for opening the issue. We're actively working on this, stay tuned :)

janekbaraniewski:
hey @lukas-mertens! We've just addedd support for signing git commits using SSH keys. Devpod will automatically set this up if your local gitconfig has specified `gpg.format=ssh` and use key specified in `user.signingkey`. Signing is NOT enforced automatically - you still need to do `git commit -S ...` in order to sign your commit. 

If you want to enable it explicitly for your workspace or if you want to use a different signing key than the one specified in your local gitconfig, just add `--git-ssh-signing-key` flag to your `devpod up` command like this:

```bash
devpod up github.com/loft-sh/devpod --git-ssh-signing-key /Users/johndoe/.ssh/mykey.pub
```

Let us know what you think!

PR #1167, opened 2024-07-15
POD-717 | Support signing commits with SSH keys 

janekbaraniewski:
This PR adds support for signing git commits inside DevPod using ssh keys.
We already support signing commits with gpg keys. This adds missing part of ssh keys support.

Issue: #1146 

# Usage

## Default way

By default DevPod will detect if user has ssh signatures set up in local git. This is done based on `git.gpg.format` git setting on users machine. If this is set to `ssh` DevPod will setup ssh signature support.

## Manual way

There is a way to manually enable ssh signatures when running `devpod up` :
```
devpod up github.com/loft-sh/devpod --git-ssh-signing-key /Users/johndoe/.ssh/mykey.pub
```


pkg/agent/tunnelserver/tunnelserver.go
```diff
@@ -216,6 +217,26 @@ func (t *tunnelServer) GitCredentials(ctx context.Context, message *tunnel.Messa
 	return &tunnel.Message{Message: string(out)}, nil
 }
 
+func (t *tunnelServer) GitSSHSignature(ctx context.Context, message *tunnel.Message) (*tunnel.Message, error) {
```
pascalbreuninger, line 220: You'll also need to implement this function in ./proxyserver (just forward the message) for Pro setups

cmd/ssh.go
```diff
@@ -472,6 +473,12 @@ func (cmd *SSHCmd) startProxyServices(
 		if gitCredentials {
 			command += " --configure-git-helper"
 		}
+
+		// check if we should enable git ssh commit signature support
+		if cmd.GitSSHSignatureForwarding || devPodConfig.ContextOption(config.ContextOptionGitSSHSignatureForwarding) == "true" {
```
pascalbreuninger, line 478: What happens if gpg agent forwarding is enabled as well?

pascalbreuninger: And by setting this we enable it for everyone even though they might not have set it up locally 🤔
Imo the intuitive way would be to enable it if it's available locally but opt-in if not. What do you think?

janekbaraniewski: yeah that's the idea, working on it atm. For testing so far it was just easier for me to have it enabled all the time. That's how I wanted to set it up:
- if you have ssh signing set up locally we'll automatically set it up for you 
- if you don't you'll be able to specify a flag with path to key you want to use for signing and we'll set it up 

And about the gpg agent forwarding - it wouldn't matter, as long as `gpg.format=ssh` is set git will use our tool. So gpg agent would be just waiting there but wouldn't be used. I'll do some testing though

PR #1183, opened 2024-07-18
fix(cli): git config path

amithm7:
ref:

- https://git-scm.com/docs/git-config/#Documentation/git-config.txt-XDGCONFIGHOMEgitconfig
- gh-1171


pascalbreuninger (approved):
LGTM, thanks!

PR #1194, opened 2024-07-30
POD-638 | Automatically auth to platform for pro users

janekbaraniewski:


ALEXIS007891076 (approved):

ALEXIS007891076 (approved):

cmd/up.go
```diff
@@ -970,6 +982,45 @@ func setupGitSSHSignature(signingKey string, client client2.BaseWorkspaceClient,
...
+	if err != nil {
+		remoteUser = "root"
+	}
+
+	port, err := credentials.GetPort()
+	if err != nil {
+		return fmt.Errorf("get port: %w", err)
+	}
+
+	command := fmt.Sprintf("devpod agent setup-loft-platform-access --context %v --provider %v --port %v", context, provider, port)
```
pascalbreuninger, line 1002: Instead of using `devpod` directly, use `client.AgentPath()`. This way it doesn't need to be available on the PATH when we ssh into the container

pascalbreuninger: Additionally, we've tried to separate the commands in terms of where things happen:
- top level -> user machine
- agent -> host machine (user for local providers, VM for cloud/ssh)
- container -> actual dev container

what do you think of moving the command to `devpod agent container setup-....`?

cmd/up.go
```diff
@@ -970,6 +982,45 @@ func setupGitSSHSignature(signingKey string, client client2.BaseWorkspaceClient,
...
 }
 
+func setupLoftPlatformAccess(context, provider string, client client2.BaseWorkspaceClient, log log.Logger) error {
+	log.Infof("Setting up Loft Platform access")
+	execPath, err := os.Executable()
+	if err != nil {
+		return err
+	}
+
+	remoteUser, err := devssh.GetUser(client.WorkspaceConfig().ID, client.WorkspaceConfig().SSHConfigPath)
```
pascalbreuninger, line 992: nit: you could use the remoteUser we already have in line 207

pkg/agent/tunnelserver/tunnelserver.go
```diff
@@ -237,6 +238,26 @@ func (t *tunnelServer) GitSSHSignature(ctx context.Context, message *tunnel.Mess
 	return &tunnel.Message{Message: string(out)}, nil
 }
 
+func (t *tunnelServer) LoftConfig(ctx context.Context, message *tunnel.Message) (*tunnel.Message, error) {
```
pascalbreuninger, line 241: I think we also need an implemenation of this in tunnelserver/proxyserver 🤔

cmd/up.go
janekbaraniewski: oh yeah definitely, that makes much more sense 👍  thx!

pkg/agent/tunnelserver/tunnelserver.go
janekbaraniewski: 👀

cmd/up.go
janekbaraniewski: `client.AgentPath` is not available here but I do have `execPath` that I can use, afaiu it's the same thing 👍

pascalbreuninger: Ah I think you can use `agent.ContainerDevPodHelperLocation` to get the correct path in the container

pascalbreuninger (approved):
LGTM 🙌

PR #1204, opened 2024-08-08
Pod 763/ expose execute timeout as context option

bkneis:
This PR fixes POD-763 by providing a new context option EXECUTE_TIMEOUT to set the timeout in seconds, instead of a hard coded value.

Additionally I have added a function to get the environment's current context options. Currently the context is loaded and it's pointer is expected to be passed round to any function that needs the config. Instead we now assign a variable during LoadConfig to act as a singleton, so it can easily be retrieved without updating lots of function signatures.

To test this PR I have run `./devpod-cli up examples/simple` to ensure injecting the binary still works, I also ran `./devpod-cli context options` to ensure the option is listed and `./devpod-cli context set-options default -o EXECUTE_TIMEOUT=60` to ensure it can be updated.

### DIFF
--- a/pkg/config/context.go
+++ b/pkg/config/context.go
@@ -13,6 +13,7 @@ const (
 	ContextOptionDotfilesScript             = "DOTFILES_SCRIPT"
 	ContextOptionSSHAgentForwarding         = "SSH_AGENT_FORWARDING"
 	ContextOptionSSHConfigPath              = "SSH_CONFIG_PATH"
+	ContextOptionExecuteTimeout             = "EXECUTE_TIMEOUT"
 )
 
 var ContextOptions = []ContextOption{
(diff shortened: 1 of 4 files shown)


pkg/config/context.go
```diff
@@ -13,6 +13,7 @@ const (
 	ContextOptionDotfilesScript             = "DOTFILES_SCRIPT"
 	ContextOptionSSHAgentForwarding         = "SSH_AGENT_FORWARDING"
 	ContextOptionSSHConfigPath              = "SSH_CONFIG_PATH"
+	ContextOptionExecuteTimeout             = "EXECUTE_TIMEOUT"
```
pascalbreuninger, line 16: Could you rename this to something that includes `agent` and `inject` to indicate which part of DevPod it's affecting? Otherwise it's hard to understand where exactly this context option comes into play 🤔

bkneis: done :)

bkneis:
@pascalbreuninger this has now been tested on GCP and seems to be working as expected

PR #1217, opened 2024-08-13
POD-763: Expose inject timeout as contect config and add to agent workspace info

bkneis:
This PR fixes POD-763 by providing a new context option AGENT_INJECT_TIMEOUT to set the timeout in seconds, instead of a hard coded value.

To test this PR I have run ./devpod-cli up examples/simple to ensure injecting the binary still works, I also ran ./devpod-cli context options to ensure the option is listed and ./devpod-cli context set-options default -o EXECUTE_TIMEOUT=40 to ensure it can be updated. Once the timeout was set I added a `sleep 60` to the inject.sh script and watched the logs to ensure the retries were separated by the new timeout.

This PR supersedes https://github.com/loft-sh/devpod/pull/1204


pkg/client/client.go
```diff
@@ -20,13 +21,13 @@ type BaseClient interface {
 	RefreshOptions(ctx context.Context, userOptions []string) error
 
 	// Status retrieves the workspace status
-	Status(ctx context.Context, options StatusOptions) (Status, error)
+	Status(ctx context.Context, options StatusOptions, cfg *config.Config) (Status, error)
```
pascalbreuninger, line 24: Looks like we need the config here in all of these functions to resolve the agent config, please correct me if I'm wrong.
If that's the case you can get a way with just resolving the value once over here: https://github.com/loft-sh/devpod/blob/c2855e63acf5bd9a8fe2fe545eee9f01392adfd0/pkg/options/resolve.go#L214
This way we wouldn't need to change the interface and all implementations for one specific line

bkneis: Thanks for pointing this out @pascalbreuninger! As it turns out the workspace client has a member variable containing the devPodConfig :disappointed: So I have simplified the "prop drilling" since lots of function now don't need this in the signature. The PR is up to date with this and is G2G in my eyes

pkg/provider/workspace.go
```diff
@@ -160,6 +161,9 @@ type AgentWorkspaceInfo struct {
 
 	// Origin holds the folder where this config was loaded from
 	Origin string `json:"-"`
+
+	// InjectTimeout specifies how long to wait for the agent to be injected into the dev container
+	InjectTimeout time.Duration `json:"inject_timeout,omitempty"`
```
pascalbreuninger, line 166: nit: all other keys are camelCase, `injectTimeout`

bkneis: yes good spot :+1:

pascalbreuninger (approved):
LGTM

PR #1228, opened 2024-08-20
fix(pro): import non-versioned workspace template

pascalbreuninger:
Fixes an issue where the import would fail if the template isn't
versioned

### DIFF
--- a/cmd/pro/import_workspace.go
+++ b/cmd/pro/import_workspace.go
@@ -163,6 +163,9 @@ func resolveInstanceOptions(ctx context.Context, instance *managementv1.DevPodWo
 	if instance.Spec.TemplateRef == nil {
 		return opts, nil
 	}
+	if instance.Spec.RunnerRef.Runner != "" {
+		opts[loft.RunnerEnv] = instance.Spec.RunnerRef.Runner
+	}
 	opts[loft.TemplateOptionEnv] = instance.Spec.TemplateRef.Name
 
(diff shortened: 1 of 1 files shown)


cmd/pro/import_workspace.go
```diff
@@ -180,9 +183,14 @@ func resolveInstanceOptions(ctx context.Context, instance *managementv1.DevPodWo
...
-	if err != nil {
-		return nil, fmt.Errorf("get template parameters: %w", err)
+	var templateParameters []storagev1.AppParameter
+	if len(template.Spec.Versions) > 0 {
+		templateParameters, err = list.GetTemplateParameters(template, instance.Spec.TemplateRef.Version)
+		if err != nil {
+			return nil, fmt.Errorf("get template parameters: %w", err)
+		}
+	} else {
+		templateParameters = template.Spec.Parameters
```
janekbaraniewski, line 193: feel free to disregard: what do you think about using this as default and overwriting with `list.GetTemplateParameters` result if `len(template.Spec.Versions) > 0`?  Or you prefer to keep declaration and initialisation of this var separate?

janekbaraniewski (approved):

cmd/pro/import_workspace.go
pascalbreuninger: that's way better, not sure what I was thinking. Ty!

Issue #1197, opened 2024-08-01
Implement `hostRequirements.gpu: "optional"`


jzazo:
**Is your feature request related to a problem?**  
Yes. When I run a devcontainer in a GPU machine I would like to make the GPUs available to the container. At the moment, I would need to add `"runArgs": ["--gpus=all"]` so the docker runtime makes the gpus available. Adding this option on a CPU only machine will turn into an error.

**Which solution do you suggest?**  
The devcontainer [specification](https://containers.dev/implementors/json_schema/) states that one can add `"hostRequirements.gpu": "optional"` and the launch command will add the `--gpus=all` argument if it detects gpus in the host. This is discussed in these issues for example [1](https://github.com/microsoft/vscode-remote-release/issues/9220#issuecomment-1820855769), [2](https://github.com/microsoft/vscode-remote-release/issues/9385).

**Which alternative solutions exist?**  
Adding `"runArgs": ["--gpus=all"]` manually on gpu nods, and removing it in cpu nodes.

bkneis:
Hi @jzazo that sounds like a great idea to support this and will make using GPUs easier, thanks for suggesting! I think we will need to add some logic here https://github.com/loft-sh/devpod/blob/main/pkg/driver/docker/docker.go#L328 that checks the dev container spec for the `hostRequirements.gpu` attribute and add `"--gpus=all"` appropriately

Benjamin-Etheredge:
This feature does not work for me. I'm not fluent in golang, but the linked MR doesn't appear to address the "optional" field being present. More logic is needed to test if the `--gpus` flag is supported, and use it if "optional" is specified.

bkneis:
@Benjamin-Etheredge thanks for reporting your issue! What exactly is not working for your use case? Can you tell me about what you are trying to accomplish? Also if you can provide your .devcontainer.json that would be ace. Are you using the docker provider?

matteosecli:
Could you please reopen?

As noted by @Benjamin-Etheredge, this feature doesn't work for `"optional"`, only for `true` or `false`.

When [`"optional"`](https://containers.dev/implementors/json_reference/#min-host-reqs) is used, DevPod should check whether the host has a working Docker+GPU setup and add `--gpus=all` only if that's the case. AFAIK, VS Code does this check by probing `docker info -f {{.Runtimes.nvidia}}`, which returns `<no value>` if the runtime is not installed.

[...]

PR #1229, opened 2024-08-20
POD-819: Add --gpus=all if host requirments contain GPU

bkneis:
This PR addresses the issue outlined in https://github.com/loft-sh/devpod/issues/1197

Effectively this PR automatically adds the appropriate `--gpus-all` flag to the run args since this is the defined behaviour by the dev container spec


pascalbreuninger (approved):
nice addition, LGTM!

PR #1241, opened 2024-09-02
Use fullDownloadURL for clarity in downloadAgentLocally function

### LINKED ISSUE #1240: Improve Logging for Download URLs in DevPod Agent
No worries, glad to hear the problem is fixed 🙂 I think the logs that mentioned this URL are referring to it as a base URL. Then depending on your platform arch it downloads the correct binary. For example `[https://github.com/loft-sh/devpod/releases/download/v0.5.19/devpod-darwin-amd64](https://github.com/loft-sh/devpod/releases/download/v0.5.19/devpod-darwin-amd64)` works fine which downloads the am64 binary

_Originally posted by @bkneis in https://github.com/loft-sh/devpod/issues/1210#issuecomment-2304334976_

Following up on this comment regarding the proxy error and URL handling:

It was mentioned that the URL in the logs is a base URL, and the actual binary downloaded depends on the platform architecture. To enhance debugging, especially in cases where network issues or incorrect URLs are involved, I suggest improving the logging to include the full constructed download URL.

This would make it easier to diagnose issues like 404 errors or proxy-related problems by providing more specific information in the logs.

JochenMan:
**Improvement: Enhanced Logging for Download URLs**

This PR improves the logging of the full download URL in the `downloadAgentLocally` function, addressing potential issues where incomplete or incorrect URLs lead to errors such as 404.

This change follows up on Issue #1240 where it was discussed that the current logging might not be sufficient for diagnosing certain network-related issues.

Closes #1240.

### DIFF
--- a/pkg/agent/inject.go
+++ b/pkg/agent/inject.go
@@ -197,7 +197,7 @@ func downloadAgentLocally(tryDownloadURL, targetArch string, log log.Logger) (st
 	}
 
 	fullDownloadURL := tryDownloadURL + "/devpod-linux-" + targetArch
-	log.Infof("Attempting to download DevPod agent from: %s", fullDownloadURL)
+	log.Debugf("Attempting to download DevPod agent from: %s", fullDownloadURL)
 
 	resp, err := devpodhttp.GetHTTPClient().Get(fullDownloadURL)
 	if err != nil {
--- a/pkg/agent/inject.go
+++ b/pkg/agent/inject.go
@@ -196,7 +196,10 @@ func downloadAgentLocally(tryDownloadURL, targetArch string, log log.Logger) (st
 		return agentPath, nil
 	}
 
-	resp, err := devpodhttp.GetHTTPClient().Get(tryDownloadURL + "/devpod-linux-" + targetArch)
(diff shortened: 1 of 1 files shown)


pkg/agent/inject.go
```diff
@@ -196,7 +196,10 @@ func downloadAgentLocally(tryDownloadURL, targetArch string, log log.Logger) (st
 		return agentPath, nil
 	}
 
-	resp, err := devpodhttp.GetHTTPClient().Get(tryDownloadURL + "/devpod-linux-" + targetArch)
+	fullDownloadURL := tryDownloadURL + "/devpod-linux-" + targetArch
+	log.Infof("Attempting to download DevPod agent from: %s", fullDownloadURL)
```
pascalbreuninger, line 200: What do you think about making this a debug log?
Imo it's not something everyone needs all of the time when creating workspace but rather when troubleshooting potential issues

pascalbreuninger:
@JochenMan thanks for the contribution! Had a question about log levels but other than that LGTM

pkg/agent/inject.go
JochenMan: Sounds good, I just pushed another commit.

pascalbreuninger (approved):

PR #1244, opened 2024-09-03
POD-854 | Decouple pulling devcontainer spec from pulling project

janekbaraniewski:
Part of POD-854

### DIFF
--- a/cmd/agent/container/setup.go
+++ b/cmd/agent/container/setup.go
@@ -160,7 +160,15 @@ func (cmd *SetupContainerCmd) Run(ctx context.Context) error {
 	}
 
 	if b, err := workspaceInfo.PullFromInsideContainer.Bool(); err == nil && b {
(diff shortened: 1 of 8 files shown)


bkneis (approved):

cmd/agent/container/setup.go
```diff
@@ -200,6 +206,74 @@ func (cmd *SetupContainerCmd) Run(ctx context.Context) error {
 	return nil
 }
 
+func cloneRepository(ctx context.Context, workspaceInfo *provider2.ContainerWorkspaceInfo, helper string, log log.Logger) error {
```
pascalbreuninger, line 209: This is shared with cmd/agent/workspace/up, right? Could we consolidate it or doesn't it make sense right now?

pascalbreuninger (approved):
LGTM

PR #1245, opened 2024-09-03
Pod 823/cache

bkneis:
This PR contains the changes to facilitate the use of remote caching via a registry to speed up build times. It supports docker and kubernetes (via kaniko) and uses the context options REGISTRY_CACHE as the registry url. Caching options are not exposed currently in order to ensure the expected functionality.

The PR was tested using the following workflow:

Without cache

1. Build examples/build                                                                                                                                        2m35s
2. Add `ghcr.io/devcontainers/features/github-cli:1` as a feature to devcontainer.json                                        2m45s
3. Add files causing the build context to change (`echo "test" > examples/build/app/test`)                                 2m52s

With cache
1. Build examples/build                                                                                                                                        2m13s
2. Add `ghcr.io/devcontainers/features/github-cli:1` as a feature to devcontainer.json                                        1m25s
3. Add files causing the build context to change (`echo "test" > examples/build/app/test`)                                       13s

As you can see step 1 is almost the same, since no cache has been made yet. Step 2 however is interesting, we saved around 50% of the build time with the cache but 1m16s was used to upload the cache back to the registry. This shows that it is important for us to either omit the `--cache-to` parameter for `up` (but not `build`), or suppress pushing of the cache manifest until the end of the command in the background once the workspace / IDE has already launched.

Also note that I am using my local docker / kind cluster and a remote registry, when the registry is closer to the devcontainer I would expect even greater savings in build time due to shorter download times.

Lastly we have step 3 that simulates a common problematic workflow, which is a user updating the build context then `up`ing a workspace. Here we see significant savings where only building the last layer is needed, instead of the entire image due to a cache miss.

EDIT: I have now implemented a boolean `ExportCache` in toggle the `--cache-to` parameter, we now only upload the cache when running `build`, not `up`. This provides `up` even faster start times of workspaces as we don't wait until the cache is uploaded.

Also note for machine providers I needed to enable the containerd snapshotter flag in the docker daemon, this was done during `initWorkspace`. For non machine providers like local docker we expect the user to enable this themselves. When trying to use the remote cache without this a WARNING will be printed from docker.

NOTE: We need to merge https://github.com/loft-sh/dockerless/pull/27 first and update the dockerless release tag in single.go

Lastly, I have updated the `CalculatePrebuildHash` function to traverse the parsed Dockerfile and extract any file paths that could affect the build context. I then filter for these files as an "include" list before adding the path to the hashed contents. This should cause fewer cache misses and allow developers to reuse similar images.


cmd/agent/workspace/up.go
```diff
@@ -206,13 +210,19 @@ func initWorkspace(ctx context.Context, cancel context.CancelFunc, workspaceInfo
...
-	errChan := make(chan error, 1)
+	errChan := make(chan error, 2)
 	go func() {
 		if !workspaceInfo.Agent.IsDockerDriver() || workspaceInfo.Agent.Docker.Install == "false" {
 			errChan <- nil
 		} else {
 			errChan <- installDocker(logger)
 		}
+		// If we are provisioning the machine, ensure the daemon has required options
+		if workspaceInfo.Machine != nil {
```
bkneis, line 221: is this the best way to check if we are using a machine provider? I only want to configure the daemon on infrastructure we provision

cmd/agent/workspace/up.go
```diff
@@ -420,11 +529,33 @@ func installDocker(log log.Logger) error {
...
+	// otherwise assume default
+	if err != nil {
+		if err = os.WriteFile("/etc/docker/daemon.json", daemonConfig, 0644); err != nil {
 			return err
 		}
 	}
-
-	return nil
+	// reload docker daemon
+	return exec.CommandContext(ctx, "pkill", "-HUP", "dockerd").Run()
```
bkneis, line 463: is this the best way to reload the docker daemon? I was surprised docker did not have a command for this and I didn't think I could assume `systemd` was being used to manage docker

janekbaraniewski: this looks scary 😬

pkg/dockerfile/parse.go
```diff
@@ -73,6 +73,24 @@ func (d *Dockerfile) FindBaseImage(buildArgs map[string]string, target string) s
 	return ""
 }
 
+// BuildContextFiles traverses a build stage and returns a list of any file path that would affect the build context
+func (d *Dockerfile) BuildContextFiles() (files []string) {
```
janekbaraniewski, line 77: some unit tests for this would be great, just to be safe

cmd/agent/workspace/up.go
janekbaraniewski: can't think of a better way to do it though, unless we'd start this build container with already mounted config? is that even possible?

janekbaraniewski:
LGTM!

cmd/agent/workspace/up.go
bkneis: I agree, maybe I could try to detect how dockerd is managed (systemd,systemctl etc) using `uname` and if not available use this as a last resort? Yes that is possible, I thought of the mounted config but this would be provider specific and need to be implemented by each machine provider, gcloud, aws, digital ocean etc. :/ At least with HUP it's a reload and not restart but it's a bit "under the hood"

pkg/dockerfile/parse.go
bkneis: good point!

cmd/agent/workspace/up.go
pascalbreuninger: You could check for docker driver and if we're on a remote machine:
`workspaceInfo.Agent.IsDockerDriver()` and `workspaceInfo.Agent.Local.Bool()`

cmd/agent/container/setup.go
```diff
@@ -302,6 +302,10 @@ func dockerlessBuild(
 	args = append(args, parseIgnorePaths(dockerlessOptions.IgnorePaths)...)
 	args = append(args, "--build-arg", "TARGETOS="+runtime.GOOS)
 	args = append(args, "--build-arg", "TARGETARCH="+runtime.GOARCH)
+	if dockerlessOptions.RegistryCache != "" {
+		log.Info("Appending registry cache to dockerless build arguments ", dockerlessOptions.RegistryCache)
```
pascalbreuninger, line 306: nit: Is this relevant for most workspace runs or could `debug` be enough?

examples/build/README.md
```diff
@@ -0,0 +1,6 @@
+## Build Example
+
+This folder holds a super simple devcontainer configuration that builds a local Dockerfile with a devcontainer feature. You can start this project via:
```
pascalbreuninger, line 3: where is the devcontainer feature though? 🤔

pkg/config/context.go
```diff
@@ -86,4 +87,9 @@ var ContextOptions = []ContextOption{
 		Description: "Specifies the timeout to inject the agent",
 		Default:     "20",
 	},
+	{
+		Name:        ContextOptionRegistryCache,
+		Description: "Specifies the registry to use as a build cache",
```
pascalbreuninger, line 92: In what format does it need to be passed in?

pkg/devcontainer/build.go
```diff
@@ -287,7 +289,7 @@ func (r *runner) buildImage(
 		return nil, err
 	}
 
-	prebuildHash, err := config.CalculatePrebuildHash(parsedConfig.Config, options.Platform, targetArch, config.GetContextPath(parsedConfig.Config), dockerfilePath, dockerfileContent, r.Log)
+	prebuildHash, err := config.CalculatePrebuildHash(parsedConfig.Config, options.Platform, targetArch, config.GetContextPath(parsedConfig.Config), dockerfilePath, dockerfileContent, r.Log, buildInfo)
```
pascalbreuninger, line 292: super small nitpick, I personally prefer the logger to be the last argument:
```suggestion
	prebuildHash, err := config.CalculatePrebuildHash(parsedConfig.Config, options.Platform, targetArch, config.GetContextPath(parsedConfig.Config), dockerfilePath, dockerfileContent, buildInfo, r.Log)
```

feel free to ignore

examples/build/README.md
bkneis: ah I was adding / removing it to test build times for caching. I must have forgot to add it back! I'll do it now

pkg/config/context.go
bkneis: good point, I'll update

cmd/agent/workspace/up.go
```diff
@@ -215,7 +215,8 @@ func initWorkspace(ctx context.Context, cancel context.CancelFunc, workspaceInfo
 			errChan <- installDocker(logger)
 		}
 		// If we are provisioning the machine, ensure the daemon has required options
-		if workspaceInfo.Machine != nil {
+		local, err := workspaceInfo.Agent.Local.Bool()
+		if workspaceInfo.Agent.IsDockerDriver() && err != nil && local {
```
pascalbreuninger, line 219: probably `!local`, no? We only want to change it on the remote instances

bkneis: :facepalm: yes your right

bkneis:
@pascalbreuninger fab thanks! Just finishing testing with a remote k8s cluster on GKE then it should be ready to merge. Do I need to update any file to reference the new version of the kubernetes driver? i.e. https://github.com/loft-sh/devpod-provider-kubernetes/pull/55

pascalbreuninger:
@bkneis nope, that's done by releasing a new version of the kubernetes provider over in the other repo👍

bkneis:
@pascalbreuninger just finished testing with GKE :partying_face: It was my auto pilot cluster giving me issues and killing workspaces with a return code 137 (OOM). Using the new GKE cluster workspaces are spinning up no problem and using the cache. In the end I didn't need to make any changes to the kubernetes driver, only dockerless so this PR is good to go IMO

PR #1253, opened 2024-09-11
POD-859 | Use Environment Templates from Pro

janekbaraniewski:
Part of POD-859


cmd/up.go
```diff
@@ -153,6 +153,7 @@ func NewUpCmd(flags *flags.GlobalFlags) *cobra.Command {
 	upCmd.Flags().StringVar(&cmd.DevContainerImage, "devcontainer-image", "", "The container image to use, this will override the devcontainer.json value in the project")
 	upCmd.Flags().StringVar(&cmd.DevContainerPath, "devcontainer-path", "", "The path to the devcontainer.json relative to the project")
 	upCmd.Flags().StringVar(&cmd.DevContainerSource, "devcontainer-source", "", "External devcontainer.json source")
+	upCmd.Flags().StringVar(&cmd.EnvironmentTemplate, "environment-template", "", "Environment template to use")
```
pascalbreuninger, line 156: could you hide that flag please? It's not expected to be used by end users directly, only in the Pro deployment

janekbaraniewski: ah good point, definitely! thx!

janekbaraniewski: done 👍

bkneis (approved):
LGTM

PR #1266, opened 2024-09-17
Log and wrap error when setting up loft platform

bkneis:
A user reported an error setting up the loft platform

https://app.usepylon.com/issues?conversationID=4a04f1ed-4a71-47e9-8b15-46edea30fdb3

Unfortunately, we didn't log or return the error so it is difficult to debug. This PR addresses the problem by wrapping the platform error and returning it.

NOTE: Is returning the error correct? I was unsure on expected behaviour, if we can't get the executable or port number we seem to return the error. But if the loft platform access error'd then we simply logged it. If we should NOT return the error I can update the PR to simply log the wrapped error.

UPDATE: I removed returning the error as this could cause more trouble than solve until we know why the bellow happens.

UPDATE: I reproduced the error locally by simply running `devpod up https://github.com/microsoft/vscode-remote-try-go`

```
08:26:38 error failure in setting up Loft Platform access: 08:26:38 info 07:26:37 fatal unknown flag: --port
08:26:38 fatal inner tunnel: Process exited with status 1
```

These are the resulting logs. I SSH'd onto the workspace and found that the entire `setup-loft-access` command was missing from `devpod agent container`.

@janekbaraniewski can you see any way that the command would not be present? I thought we scp'd the devpod binary over the container, it's odd my local devpod had the command but on the workspace not.

### DIFF
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -254,7 +254,6 @@ func (cmd *UpCmd) Run(
 	if cmd.SetupLoftPlatformAccess {
 		err = setupLoftPlatformAccess(devPodConfig.DefaultContext, context.DefaultProvider, user, client, log)
 		if err != nil {
-			log.Error(err)
 			return err
 		}
 	}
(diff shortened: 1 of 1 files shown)


janekbaraniewski (approved):

janekbaraniewski:
LGTM, as long as there is no reason we didn't error there in the first place 😅

PR #1270, opened 2024-09-18
fix: fix wrong parameter order

cuishuang:
There is an example:

```go
package main

import (
	"fmt"
	"strings"
)

func main() {

	cmd := struct {
		InstallScript string
	}{
		InstallScript: "./example/script.sh",
	}

	commandOld := "./" + strings.TrimPrefix("./", cmd.InstallScript)
	fmt.Println("old Constructed command:", commandOld)

	commandNew := "./" + strings.TrimPrefix(cmd.InstallScript, "./")
	fmt.Println("new Constructed command:", commandNew)
}

```

Output:

```
old Constructed command: ././
new Constructed command: ./example/script.sh
```

https://go.dev/play/p/OpHgx2LHtsp


pascalbreuninger (approved):
LGTM, thanks 🙌

Issue #1284, opened 2024-09-27
Permission error when using podman and having different uid between host and remote users


aacebedo:
**What happened?**  
There is an issue with uid and permissions settings when using podman. It is due to the addition of the `userns=keep-id` option that is now added automatically by devpod when podman is used (wihich is a good thing IMHO).

This option makes podman to recopy the host user inside the container with the same uid. 
If one create a user inside the container with a uid different than the host user, devpod will not be able to change the uid of the remoteUser as the host user now already exists in the container.

However, devpod still chown everything with the initial remoteUser uid which was not changed and thus breaks all permissions on the host.

This behaviour is not exhibited when uid of host user and remote user are the same. In this case, podman does not create the user in the container and keep the already existing remoteuser. Devpod does not have anything to change then.

**What did you expect to happen instead?**  
Devpod shall be able to change the uid of the remoteUser even if a user already exists.

**How can we reproduce the bug?** (as minimally and precisely as possible)  

My *`devcontainer.json`*:
```
{
  "image": "ubuntu",
  "remoteUser": "johndoe",
  "features": {
      "./test":{}
  },
}
```

test feature:
```
#!/bin/bash
set -e

export DEBIAN_FRONTEND=noninteractive

OS_ID=$(awk -F "=" '/^ID=/ {print $2}' /etc/os-release)

echo "export TZ=Europe/Paris" >> /etc/profile
echo "export LANG=en_US.UTF-8" >> /etc/profile
echo "export LANGUAGE=en_US:en" >> /etc/profile
echo "export LC_ALL=en_US.UTF-8" >> /etc/profile
echo "export LC_CTYPE=en_US.UTF-8" >> /etc/profile

source /etc/profile
apt update
apt install -y --no-install-recommends \
                sudo \
                locales \
                ssh
sed -i '/en_US.UTF-8/s/^# //g' /etc/locale.gen
locale-gen

useradd -u 10000 -m -N  -g users johndoe
passwd --delete johndoe
usermod -aG sudo johndoe
```

Ensure that host user does not have uid 10000

**Local Environment:**  
- DevPod Version: 0.5.20
- Operating System:  linux
- ARCH of the OS: AMD64 

**DevPod Provider:**  
- Local/remote provider: docker with podman

**Anything else we need to know?**

aacebedo:
In fact I checked the code and I think that `UpdateRemoteUserUID` is not used by devpod. So permissions uid is never changed. 
[...]

PR #1287, opened 2024-09-27
Implement UpdateRemoteUserID handling

aacebedo:
This PR is a tentative implementation of the UpdateRemoteUserID handling.

The reference implementation uses an additional dockerfile (https://github.com/devcontainers/cli/blob/main/scripts/updateUID.Dockerfile) to create a new image using the container image created previously.

I don't think this approach works well with prebuilding feature of devpod. So I decided to amend the
container after it has been created. 

In some case, this modification may make the same uid gid to be used twice for 2 users and 2 groups but in the end it shall not impact the way the container work in the end. 

This also solves #1284.

### DIFF
--- a/pkg/driver/docker/docker.go
+++ b/pkg/driver/docker/docker.go
@@ -1,10 +1,13 @@
 package docker
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"io"
 	"os"
+	"os/user"
+	"regexp"
 	"runtime"
 	"slices"
 	"strconv"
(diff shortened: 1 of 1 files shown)


pascalbreuninger:
@aacebedo Thanks for the PR! Let's run the test suite real quick and see if it works out.
We've been working on this a while ago as well, although without using the `UpdateRemoteUserID` option. See [this PR](https://github.com/loft-sh/devpod/pull/551/files). 
At the time we've abandonded it because of potential breaking changes and we'd need to invest more time figuring out the consequences. 

As for your PR I think scoping it only to linux, as per the spec, would already limit the potential blast radius. I'll still need to test the change with a number of provider/workspace combinations

aacebedo:
fixed the lint issue

pkg/driver/docker/docker.go
```diff
@@ -363,6 +366,188 @@ func (d *dockerDriver) RunDockerDevContainer(
...
+		if err != nil {
+			return err
+		}
+		localUid := localUser.Uid
+		localGid := localUser.Gid
+
+		// Retrieve user to update
+		containerUser := parsedConfig.RemoteUser
+		if containerUser == "" {
+			containerUser = parsedConfig.ContainerUser
```
bkneis, line 382: do we need to check if `parsedConfig.ContainerUser` is ""? If so should we default to root?

pkg/driver/docker/docker.go
```diff
@@ -363,6 +366,188 @@ func (d *dockerDriver) RunDockerDevContainer(
...
+		}
+		containerPasswdFileOut, err := os.CreateTemp("", "devpod_container_passwd_out")
+		if err != nil {
+			return err
+		}
+		containerGroupFileOut, err := os.CreateTemp("", "devpod_container_group_out")
+		if err != nil {
+			return err
+		}
+		defer os.Remove(containerPasswdFileIn.Name())
```
bkneis, line 408: each defer needs to proceed the file creation, otherwise the last file could fail and we would have the first 4 not cleaned up

pkg/driver/docker/docker.go
```diff
@@ -363,6 +366,188 @@ func (d *dockerDriver) RunDockerDevContainer(
...
+		defer os.Remove(containerGroupFileIn.Name())
+		defer os.Remove(containerPasswdFileOut.Name())
+		defer os.Remove(containerGroupFileOut.Name())
+
+		// Copy /etc/passwd and /etc/group from the container to the temporary files
+		{
```
bkneis, line 414: why are these two operations scoped to anonymous blocks?

pkg/driver/docker/docker.go
```diff
@@ -363,6 +366,188 @@ func (d *dockerDriver) RunDockerDevContainer(
...
+			args := []string{"cp", fmt.Sprintf("%s:/etc/group", container.ID), containerGroupFileIn.Name()}
+			d.Log.Debugf("Running docker command: %s %s", d.Docker.DockerCommand, strings.Join(args, " "))
+			err = d.Docker.Run(ctx, args, nil, writer, writer)
+			if err != nil {
+				return err
+			}
+		}
+
+		containerPasswdFileIn, err = os.Open(containerPasswdFileIn.Name())
```
bkneis, line 432: needs a `defer containerPasswdFileIn.Close()`

pkg/driver/docker/docker.go
```diff
@@ -363,6 +366,188 @@ func (d *dockerDriver) RunDockerDevContainer(
...
+
+		if err := scanner.Err(); err != nil {
+			return err
+		}
+
+		if localUid == "0" || containerUid == "0" || (localUid == containerUid && localGid == containerGid) {
+			return nil
+		}
+
+		containerGroupFileIn, err = os.Open(containerGroupFileIn.Name())
```
bkneis, line 477: needs closing :)

pkg/driver/docker/docker.go
```diff
@@ -363,6 +366,188 @@ func (d *dockerDriver) RunDockerDevContainer(
...
+			return err
+		}
+
+		// Update /etc/passwd and /etc/group with the new user UID and GID
+		scanner := bufio.NewScanner(containerPasswdFileIn)
+		containerUid := ""
+		containerGid := ""
+		containerHome := ""
+
+		re := regexp.MustCompile(fmt.Sprintf(`^%s:(?P<password>x?):(?P<uid>.*):(?P<gid>.*):(?P<gcos>.*):(?P<home>.*):(?P<shell>.*)$`, containerUser))
```
bkneis, line 443: why are we parsing the `/etc/group` and `/etc/passwd` file and rewriting it? What prevents us from simply copying the file over?

pkg/driver/docker/docker.go
aacebedo: I don't think so because we would change the UID of root. I don't think it is a good thing.
According to spec it is only applicable when remoteUser or containerUser is set. So maybe I shall return in this case directly.

pkg/driver/docker/docker.go
aacebedo: Agreed. Changed it.

pkg/driver/docker/docker.go
aacebedo: Right. Corrected

pkg/driver/docker/docker.go
aacebedo: I did not find a way to write in place in a file in go. You know how to do it?

pkg/driver/docker/docker.go
aacebedo: yup corrected :)

pkg/driver/docker/docker.go
aacebedo: Artifact of previous version. Removed it.

pkg/driver/docker/docker.go
bkneis: Yes for sure let's not do that :) if the spec states it is only for remote or container user then I think we should return. There are definitely use cases where neither are set

pkg/driver/docker/docker.go
bkneis: apologies I misunderstood the code. We are mutating the file, since we override `containerUser` based on the regex failing. So I think the implementation is sound

pkg/driver/docker/docker.go
```diff
@@ -363,6 +366,191 @@ func (d *dockerDriver) RunDockerDevContainer(
...
+		}
+
+		if err := scanner.Err(); err != nil {
+			return err
+		}
+
+		d.Log.Debugf("Updating container user uid")
+
+		// Copy /etc/passwd and /etc/group back to the container
+		{
```
bkneis, line 517: also don't think we need these anonymous blocks

aacebedo: corrected

aacebedo:
If everything is OK can we merge?

aacebedo:

pascalbreuninger:
@aacebedo I haven't forgotten this PR, I'm still thinking through it and weigh against other options. Thanks for the work you've put in, bear with me :)

aacebedo:
Thanks!

I have some users who are facing the issue and it is very annoying for them.

Do you have an idea of an ETA? Can I help?

bkneis:
@aacebedo I've tested this locally but did not get the result I thought I would. Using the following devcontainer.json and Dockerfile, I toggled `updateRemoteUserUID` from true to false. Here are the resulting passwd files

```
{
  "name": "Basic Python",
  "build": { "dockerfile": "Dockerfile" },
  "remoteUser": "vscode",
  "containerUser": "vscode",
  "updateRemoteUserUID": false,
  "customizations": {
    "vscode": {
      "extensions": ["ms-python.python"]
    }
  }
}

```
FROM library/python:3.11-bookworm

ARG USER=vscode
ARG DEBIAN_FRONTEND=noninteractive
RUN apt update \
    && apt install -y --no-install-recommends sudo \
    && apt autoremove -y \
    && rm -rf /var/lib/apt/lists/* \
    && useradd -m -s /usr/bin/bash ${USER} \
    && echo "${USER} ALL=(ALL) NOPASSWD: ALL" >/etc/sudoers.d/${USER} \
    && chmod 0440 /etc/sudoers.d/${USER}

```

When updateRemoteUserUID is true
```
root:x:0:
daemon:x:1:
...
vscode:x:1000:
```

When updateRemoteUserUID is false
```
root:x:0:0:root:/root:/bin/bash
daemon:x:1:1:daemon:/usr/sbin:/usr/sbin/nologin
...
vscode:x:1000:1000::/home/vscode:/usr/bin/bash
```

Should the false entry have dropped the extra info? Also what is the current issue you are facing? When I set this to false, I thought I would have permission issues but can still mount my workspace

aacebedo:
> @aacebedo I've tested this locally but did not get the result I thought I would. Using the following devcontainer.json and Dockerfile, I toggled `updateRemoteUserUID` from true to false. Here are the resulting passwd files
> 
> ```
> {
>   "name": "Basic Python",
>   "build": { "dockerfile": "Dockerfile" },
>   "remoteUser": "vscode",
>   "containerUser": "vscode",
>   "updateRemoteUserUID": false,
>   "customizations": {
>     "vscode": {
>       "extensions": ["ms-python.python"]
>     }
>   }
> }
> ```
> 
> ```
> FROM library/python:3.11-bookworm
> 
> ARG USER=vscode
> ARG DEBIAN_FRONTEND=noninteractive
> RUN apt update \
>     && apt install -y --no-install-recommends sudo \
>     && apt autoremove -y \
>     && rm -rf /var/lib/apt/lists/* \
>     && useradd -m -s /usr/bin/bash ${USER} \
>     && echo "${USER} ALL=(ALL) NOPASSWD: ALL" >/etc/sudoers.d/${USER} \
>     && chmod 0440 /etc/sudoers.d/${USER}
> ```
> 
> When updateRemoteUserUID is true
> 
> ```
> root:x:0:
> daemon:x:1:
> ...
> vscode:x:1000:
> ```
> 
> When updateRemoteUserUID is false
> 
> ```
> root:x:0:0:root:/root:/bin/bash
> daemon:x:1:1:daemon:/usr/sbin:/usr/sbin/nologin
> ...
> vscode:x:1000:1000::/home/vscode:/usr/bin/bash
> ```
> 
> Should the false entry have dropped the extra info? Also what is the current issue you are facing? When I set this to false, I thought I would have permission issues but can still mount my workspace

Hi @bkneis 

You printed the /etc/passwd of the container right? I'll check

For the issue I am facing, it is described in+ #1284

bkneis:
@aacebedo Thanks for clarifying. Yes these two files were from the container. My host user has UID 1000 so I was expecting a different UID when setting to true. I am not using podman, but I don't think this should make a difference for this test? As it should still parse my /etc/passwd and check the UID? Just want to make sure the PR is working as expected before merging

aacebedo:
@bkneis could you give me more details about the commands to reproduce?

aacebedo:
> @aacebedo I just retested with your branch but could not reproduce my issue. Toggling `updateRemoteUserUID` both have
> 
> ```
> vscode@0c8cb50a3aa3:/workspaces/ac4$ cat /etc/passwd
> root:x:0:0:root:/root:/bin/bash
> daemon:x:1:1:daemon:/usr/sbin:/usr/sbin/nologin
> bin:x:2:2:bin:/bin:/usr/sbin/nologin
> sys:x:3:3:sys:/dev:/usr/sbin/nologin
> sync:x:4:65534:sync:/bin:/bin/sync
> games:x:5:60:games:/usr/games:/usr/sbin/nologin
> man:x:6:12:man:/var/cache/man:/usr/sbin/nologin
> lp:x:7:7:lp:/var/spool/lpd:/usr/sbin/nologin
> mail:x:8:8:mail:/var/mail:/usr/sbin/nologin
> news:x:9:9:news:/var/spool/news:/usr/sbin/nologin
> uucp:x:10:10:uucp:/var/spool/uucp:/usr/sbin/nologin
> proxy:x:13:13:proxy:/bin:/usr/sbin/nologin
> www-data:x:33:33:www-data:/var/www:/usr/sbin/nologin
> backup:x:34:34:backup:/var/backups:/usr/sbin/nologin
> list:x:38:38:Mailing List Manager:/var/list:/usr/sbin/nologin
> irc:x:39:39:ircd:/run/ircd:/usr/sbin/nologin
> _apt:x:42:65534::/nonexistent:/usr/sbin/nologin
> nobody:x:65534:65534:nobody:/nonexistent:/usr/sbin/nologin
> vscode:x:1000:1000::/home/vscode:/usr/bin/bash
> ```
> 
> when the UID is the same, which is expected. So it must have been finger trouble my end

Great so we can pass the tests on it and merge it then?

aacebedo:
Rebased on main

pascalbreuninger (approved):

PR #1288, opened 2024-09-27
Add send-env and set-env options to ssh command

aacebedo:
Having similar options as SetEnv and SendEnv from the ssh client can be useful.

This PR adds those options to the ssh command.

### DIFF
--- a/cmd/ssh.go
+++ b/cmd/ssh.go
@@ -40,6 +40,8 @@ type SSHCmd struct {
 	ForwardPortsTimeout string
 	ForwardPorts        []string
 	ReverseForwardPorts []string
+	SendEnvVars         []string
+	SetEnvVars          []string
 
 	Stdio                     bool
 	JumpContainer             bool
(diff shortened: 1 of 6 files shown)


pascalbreuninger:
Hey @aacebedo, thanks for the PR!
I'm wondering how this differs from the `--workspace-env` flag in `devpod up`, except that you can set env vars per-connection.
 
Do you have a specific use case in mind?

aacebedo:
My UC is the following.
I start a container during a CI step through a common template and I don't know before being actually used in the step which variables will be used.

With this, I am able to inject directly the variable I need when I write it.
In addition having temporary variables valid for a single command seems better than having to create a new container for each command (when using ssh --command option)

bkneis:
LGTM, both options are supported by SSH so I don't see why we shouldn't support

bkneis (approved):

cmd/ssh.go
```diff
@@ -38,6 +38,8 @@ type SSHCmd struct {
 	ForwardPortsTimeout string
 	ForwardPorts        []string
 	ReverseForwardPorts []string
+	SentEnvVars         []string
```
pascalbreuninger, line 41: small nit: 
```suggestion
	SendEnvVars         []string
```

pkg/ssh/helper.go
```diff
@@ -90,13 +90,22 @@ func ConfigFromKeyBytes(keyBytes []byte) (*ssh.ClientConfig, error) {
 	return clientConfig, nil
 }
 
-func Run(ctx context.Context, client *ssh.Client, command string, stdin io.Reader, stdout io.Writer, stderr io.Writer) error {
+func Run(ctx context.Context, client *ssh.Client, command string, stdin io.Reader, stdout io.Writer, stderr io.Writer, envVars *map[string]string) error {
```
pascalbreuninger, line 93: why does this have to be a pointer? The default value of `map[string]string` is already `nil`

cmd/ssh.go
aacebedo: Corrected

pkg/ssh/helper.go
aacebedo: corrected

aacebedo:
I fixed the comment. Will it be possible te review please?

aacebedo:

pascalbreuninger:
@aacebedo merged!

PR #1298, opened 2024-10-03
POD-944 | Pass provided git credentials to crane

janekbaraniewski:
Part of POD-944


bkneis (approved):

cmd/up.go
```diff
@@ -425,6 +425,8 @@ func (cmd *UpCmd) devPodUpProxy(
 		baseOptions.IDE = workspace.IDE.Name
 		baseOptions.IDEOptions = nil
 		baseOptions.Source = workspace.Source.String()
+		baseOptions.GitOverrideUsername = cmd.GitUsername
```
pascalbreuninger, line 428: Not sure if we need this here. This will be executed on the dev machine and they won't have the username and token flag set

pkg/provider/workspace.go
```diff
@@ -204,6 +204,8 @@ type CLIOptions struct {
 	GitCloneStrategy     git.CloneStrategy `json:"gitCloneStrategy,omitempty"`
 	FallbackImage        string            `json:"fallbackImage,omitempty"`
 	GitSSHSigningKey     string            `json:"gitSshSigningKey,omitempty"`
+	GitOverrideUsername  string            `json:"gitUsername,omitempty"`
```
pascalbreuninger, line 207: If we have it on the CLIOptions already, do we still need it on this struct?

cmd/up.go
janekbaraniewski: this is more for testing, since those args are hidden anyway. can remove though if you think it's misleading

pkg/provider/workspace.go
janekbaraniewski: currently we have it set on UpCmd struct and not workspace CLIOptions

cmd/up.go
pascalbreuninger: it's a bit unintuitive when reading through it but `devPodUpProxy` will be executed on the devs machine, initiating the websocket connection and `devPodUpMachine` will then be run on the remote side

pkg/provider/workspace.go
pascalbreuninger: whoopsies, my bad! thanks

cmd/up.go
janekbaraniewski: ah 🤦  kk thx on it!

cmd/up.go
```diff
@@ -65,9 +66,9 @@ type UpCmd struct {
 }
 
 // NewUpCmd creates a new up command
-func NewUpCmd(flags *flags.GlobalFlags) *cobra.Command {
+func NewUpCmd(f *flags.GlobalFlags) *cobra.Command {
 	cmd := &UpCmd{
-		GlobalFlags: flags,
+		GlobalFlags: f,
```
pascalbreuninger, line 71: probably neeeds type embedding of `dpFlags`?

Issue #1292, opened 2024-10-01
Other property of mounts should be serialized


gjhenrique:
**What happened?**  
I tried to use the `other` property in the `mounts`

```json
{
  "mounts": [
    {
      "source": "/home/host/filei",
      "target": "/home/container/fili",
      "type": "bind",
      "other": ["readonly"]
    }
  ],
}
```

**What did you expect to happen instead?**  
The mount should be read-only

**How can we reproduce the bug?** (as minimally and precisely as possible)  

Add `other` as mounts and check if the value is passed to `docker run` 

Actually, the bug happens because there is no treatment for `other` property. I was going insane trying to find why go wasn't marshaling the struct properly, but the issue was 

It worked with this code inside the [UnmarshalJSON](https://github.com/loft-sh/devpod/blob/84a4d0179b3fbe01b9a1d785f08ed3dbf860c38b/pkg/devcontainer/config/config.go#L411-L427) method

```golang
Unstaged changes (2)
modified   pkg/devcontainer/config/config.go
@@ -424,6 +424,16 @@ func (m *Mount) UnmarshalJSON(data []byte) error {
 		if ok {
 			m.External = externalStr
 		}
+
+		valInterface, ok := obj["other"].([]interface{})
+		if ok {
+			other := make([]string, len(valInterface))
+			for i, _ := range valInterface {
+				other[i] = valInterface[i].(string)
+			}
+			m.Other = other
+		}
```

**Local Environment:**  
- DevPod Version: v0.5.20
- Operating System: linux
- ARCH of the OS: AMD64

**DevPod Provider:**  
- Local/remote provider: docker 

**Anything else we need to know?**  
Not really. I can also open the PR if it's faster, since it works on my machine :tm:

bkneis:
@gjhenrique thanks for reporting the issue! Looking at the code I can see your fix looks good to me. Can you please place a PR with these changes?

PR #1306, opened 2024-10-07
Support parsing "other" property from mounts

gjhenrique:
Fixes https://github.com/loft-sh/devpod/issues/1292


pascalbreuninger:
@gjhenrique thanks for the contribution, I can't believe we've missed that!

PR #1313, opened 2024-10-12
Replace direct use of su with sudo su

nrontsis:
To avoid complications when running as non-root, as discussed in https://github.com/loft-sh/devpod-provider-kubernetes/issues/12

### DIFF
--- a/pkg/devcontainer/setup/lifecyclehooks.go
+++ b/pkg/devcontainer/setup/lifecyclehooks.go
@@ -98,7 +98,7 @@ func run(commands []types.LifecycleHook, user, dir string, remoteEnv map[string]
 			log.Infof("Run command %s: %s...", k, strings.Join(c, " "))
 			args := []string{}
 			if user != "root" {
-				args = append(args, "su", user, "-c", command.Quote(c))
+				args = append(args, "sudo", "su", user, "-c", command.Quote(c))
 			} else {
 				args = append(args, "sh", "-c", command.Quote(c))
 			}


bkneis:
@nrontsis thanks for clarifying, I understand your use case much better now :) What you suggested seems better, but I still have concerns that it could escalate privilege to a user that should not be a sudo user. Perhaps something like this?

nrontsis:
> What you suggested seems better, but I still have concerns that it could escalate privilege to a user that should not be a sudo user.

Just to clarify what I am now proposing reflected [this new PR](https://github.com/loft-sh/devpod/pull/1317/files). In that case I can't image how privilege escalation could happen, as I am not using `sudo` there, but direct `su` instead, which would require root's password.

> Also @nrontsis can you explain how you tested this PR and if it worked for your use case?

Currently testing end-to-end in a private repo I am working on. I am happy to create a minimal example that demonstrates this working, if you are conceptually happy with the appoach.

sanmai-NL:
Please don't check for a username, check for a UID.

bkneis:
@sanmai-NL he's comparing 2 User structs, which will contain UID and username :) @nrontsis I like the implementation of 1317. If this works for you, can you please post a summary of your use case / test / working example on that PR

nrontsis:
great, closing this PR and continue there.

nrontsis:
@bkneis I updated #1317 with a description of the use case and added a conceptual minimal example. I did not test the minimal example provided end to end, I was hoping that the change is small enough that this might not be necessary, but let me know if you think otherwise.

PR #1317, opened 2024-10-16
Don't hardcode root when running lifecycle scripts

nrontsis:
As discussed in https://github.com/loft-sh/devpod/pull/1313#issuecomment-2416787129, this PR avoids a hardcoded assumption from the lifecycle scripts that the devpod is built under a root user.

Building the devpod under a non-root user can be important in restricted kubernetes corporate environments. While currently creating devpods under a non-root user in kubernetes requires [some further hacks](https://github.com/nrontsis/devpod-example-go/compare/63cc83c...main), this PR removes a blocking certain aspect of it, allowing for creating a devpod as per the following demo. Before this PR, the building of the devpod would stop because `"su", user, "-c", command.Quote(c)` would be called by the run lifecycle hooks logic

### Minimal example (not tested end-to-end)

```bash
devpod provider add kubernetes k8s-demo
devpod provider set-options k8s-demo \
  --option STRICT_SECURITY=true \
  --option POD_MANIFEST_TEMPLATE="$(cat k8s_template.yaml)"
devpod up git@github.com:nrontsis/devpod-example-go.git --ide=none --provider=k8s-demo --open-ide=false
```

where `k8s_template.yaml` is
```yaml
kind: Pod
apiVersion: v1
spec:
  initContainers:
  - name: set-volume-owner
    image: busybox
    command: [ "sh", "-c", "chown -R 1000:1000 /volume" ]
    volumeMounts:
    - mountPath: /volume
      name: devpod
      subPath: devpod/0
  containers:
    - name: devpod
      securityContext: 
        runAsGroup: 1000
        runAsUser: 1000
        runAsNonRoot: true
status: {}
```

### DIFF
--- a/pkg/devcontainer/setup/lifecyclehooks.go
+++ b/pkg/devcontainer/setup/lifecyclehooks.go
@@ -70,7 +70,7 @@ func RunLifecycleHooks(ctx context.Context, setupInfo *config.Result, log log.Lo
 	return nil
 }
 
-func run(commands []types.LifecycleHook, user, dir string, remoteEnv map[string]string, name, content string, log log.Logger) error {
+func run(commands []types.LifecycleHook, remoteUser, dir string, remoteEnv map[string]string, name, content string, log log.Logger) error {
 	if len(commands) == 0 {
 		return nil
 	}
@@ -102,8 +102,8 @@ func run(commands []types.LifecycleHook, user, dir string, remoteEnv map[string]
 				log.Fatalf("Error fetching the current user: %v", err)
 			}
 			args := []string{}
-			if user != currentUser {
-				args = append(args, "su", user, "-c", command.Quote(c))
+			if remoteUser != currentUser.Username {
+				args = append(args, "su", remoteUser, "-c", command.Quote(c))
 			} else {
 				args = append(args, "sh", "-c", command.Quote(c))
 			}
--- a/pkg/devcontainer/setup/lifecyclehooks.go
+++ b/pkg/devcontainer/setup/lifecyclehooks.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"os/user"
 	"regexp"
 	"slices"
 	"strings"
@@ -96,8 +97,12 @@ func run(commands []types.LifecycleHook, user, dir string, remoteEnv map[string]
 
 		for k, c := range cmd {
 			log.Infof("Run command %s: %s...", k, strings.Join(c, " "))
+			currentUser, err := user.Current()
+			if err != nil {
+				log.Fatalf("Error fetching the current user: %v", err)
+			}
 			args := []string{}
-			if user != "root" {
+			if user != currentUser {
 				args = append(args, "su", user, "-c", command.Quote(c))
 			} else {
 				args = append(args, "sh", "-c", command.Quote(c))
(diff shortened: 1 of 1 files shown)


pkg/devcontainer/setup/lifecyclehooks.go
```diff
@@ -96,9 +97,13 @@ func run(commands []types.LifecycleHook, user, dir string, remoteEnv map[string]
 
 		for k, c := range cmd {
 			log.Infof("Run command %s: %s...", k, strings.Join(c, " "))
+			currentUser, err := user.Current()
+			if err != nil {
+				log.Fatalf("Error fetching the current user: %v", err)
```
pascalbreuninger, line 102: Can you return the error instead of panicing so we can gracefully shutdown the reset of the app?

pascalbreuninger (commented):
Had a remark about error handling, other than that LGTM. Thanks for the contribution 🙌

bkneis:
@nrontsis thanks for the contribution! The implementation is a small change with a big impact, perfect! If you have some time I would love to know an full e2e test was run, even manually, just to ensure the workspace starts with no errors and that it doesn't cause other issues

pkg/devcontainer/setup/lifecyclehooks.go
```diff
@@ -96,9 +97,13 @@ func run(commands []types.LifecycleHook, user, dir string, remoteEnv map[string]
...
+			}
 			args := []string{}
-			if user != "root" {
-				args = append(args, "su", user, "-c", command.Quote(c))
+			if remoteUser != currentUser.Username {
```
bkneis, line 105: oh we are comparing usernames, I thought we compared User structs. Is it possible to update the function signature to pass the actual remoteUser struct to `run` and compare UID instead of username

nrontsis: I think this won't be easy because the `remoteUser` value here comes from the `GetRemoteUser` function. `GetRemoteUser` returns the user name and not the id, partially also because it relies on  devcontainer.json's `remoteUser` - which specifies the user name, and not the user id.

pkg/devcontainer/setup/lifecyclehooks.go
nrontsis: Done, thanks

nrontsis:
@bkneis I've now successfully tested the minimal example end-to-end using the following steps.

First go to [my test repo](https://github.com/nrontsis/devpod-example-go/), build `.devcontainer/Dockerfile` and push i the resulting image in a registry that the k8s cluster will have access to. After that replace `.devcontainer/devcontainer.json:image` to point to that pushed image.

Then clone the patched `devpod` of this PR and do:
```
./hack/rebuild.sh
test/devpod-cli-linux-amd64 provider add kubernetes --name k8s-demo
test/devpod-cli-linux-amd64 provider set-options k8s-demo \
  --option STRICT_SECURITY=true \
  --option POD_MANIFEST_TEMPLATE="$(cat k8s_template.yaml)" 
test/devpod-cli-linux-amd64 up git@github.com:nrontsis/devpod-example-go.git --ide=none --provider=k8s-demo --open-ide=false --debug
```
where `k8s_template.yaml` is
```
kind: Pod
apiVersion: v1
spec:
  initContainers:
  - name: set-volume-owner
    image: busybox
    command: [ "sh", "-c", "chown -R 1000:1000 /volume" ]
    volumeMounts:
    - mountPath: /volume
      name: devpod
      subPath: devpod/0
  containers:
    - name: devpod
      securityContext: 
        runAsGroup: 1000
        runAsUser: 1000
        runAsNonRoot: true
status: {}
```
which runs without issues.

nrontsis:
I believe this is ready now.

bkneis (approved):
Thanks for the contribution! LGTM

nrontsis:
Thanks for the reviews! Happy to merge this?

PR #1332, opened 2024-10-23
Fix/gpg agent forwarding arg list

pascalbreuninger:
In cases where the gpg public key exceeds ARG_MAX of the workspace
shelll gpg forwarding fails because the shell rejects the command.
We now fetch the private key through the workspace credentials server
before decoding and passing to `gpg`


cmd/agent/workspace/setup_gpg.go
```diff
@@ -53,8 +53,14 @@ func NewSetupGPGCmd(flags *flags.GlobalFlags) *cobra.Command {
 func (cmd *SetupGPGCmd) Run(ctx context.Context, log log.Logger) error {
 	log.Debugf("Initializing gpg-agent forwarding")
 
-	log.Debugf("Decoding input public key")
-	publicKey, err := base64.StdEncoding.DecodeString(cmd.PublicKey)
+	log.Debugf("Fetching public key")
+	rawPublicKeys, err := getPublicKeys(log)
+	if err != nil {
+		return err
```
bkneis, line 59: should we be logging this error? also on line 65

pkg/agent/tunnelserver/tunnelserver.go
```diff
@@ -264,6 +266,17 @@ func (t *tunnelServer) LoftConfig(ctx context.Context, message *tunnel.Message)
 	return &tunnel.Message{Message: string(out)}, nil
 }
 
+func (t *tunnelServer) GPGPublicKeys(ctx context.Context, message *tunnel.Message) (*tunnel.Message, error) {
```
bkneis, line 269: docblock explaining what this function is doing would be nice

pkg/credentials/request.go
```diff
@@ -0,0 +1,59 @@
...
+
+		raw, err := io.ReadAll(response.Body)
+		if err != nil {
+			return err
+		}
+
+		// has the request succeeded?
+		if response.StatusCode != http.StatusOK {
+			log.Errorf("Error calling %s (%d): %w", endpoint, response.StatusCode, string(raw))
+			return err
```
bkneis, line 47: err might be nil here, maybe assing the error above in the log then log it and return

pkg/credentials/request.go
```diff
@@ -0,0 +1,59 @@
...
+		}
+
+		out = raw
```
bkneis, line 50: you could simply reuse out instead of temp var raw?

bkneis (approved):
LGTM, just a couple small comments

pkg/agent/tunnelserver/tunnelserver.go
pascalbreuninger: Let's do that another time when we document all of the GRPC endpoints :)

pkg/credentials/request.go
pascalbreuninger: good catch, thanks. Err will always be nil here so we now just return a new one

pascalbreuninger:
Abandonding in favor of #1336

PR #1337, opened 2024-10-28
Implemement .devpodignore

### LINKED ISSUE #1299: Create .devpodignore to avoid copying files/folders to the devcontianer
When using kubernetes provider, source files from the local folder are copied to the remote devcontainer deployed in a pod within the cluster.

If I build the app in local, I generate the output folder with the binary and many other things that are not needed to be copied in the devcontianer. However, If I start the container, those files/folders are also copied.

In my case it was ~1GB it was copied being very slow.

Add a `.devpodignore` where you can define which files/folder to ignore when copying the source code.

bkneis:
This PR implements the logic to support `.devpodignore` file. It has been tested using the docker and kubernetes provider with a local and git workspace source. I managed to exclude a test file from a [repo with .devpodignore](https://github.com/loft-sh/devpod-example-devops)

Fixes https://github.com/loft-sh/devpod/issues/1299

### DIFF
--- a/pkg/agent/workspace.go
+++ b/pkg/agent/workspace.go
@@ -17,6 +17,7 @@ import (
 	provider2 "github.com/loft-sh/devpod/pkg/provider"
 	"github.com/loft-sh/log"
 	"github.com/mitchellh/go-homedir"
+	"github.com/moby/buildkit/frontend/dockerfile/dockerignore"
 )
 
 var extraSearchLocations = []string{"/home/devpod/.devpod/agent", "/opt/devpod/agent", "/var/lib/devpod/agent", "/var/devpod/agent"}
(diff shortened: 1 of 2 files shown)


pkg/agent/workspace.go
```diff
@@ -303,6 +304,22 @@ func CloneRepositoryForWorkspace(
...
+		return nil
+	}
+	excludes, err := dockerignore.ReadAll(f)
+	if err != nil {
+		log.Warn(".devpodignore file is invalid : ", err)
+		return nil
+	}
+	// Remove files from workspace content folder
+	for _, exclude := range excludes {
+		os.RemoveAll(filepath.Join(workspaceDir, exclude))
```
pascalbreuninger, line 319: I'm not sure if this is expected. We're using a bind mount for the devcontainer with local docker and this would delete the excludes on the users machine

bkneis: True, but only from the content folder, i.e. `~/.devpod/context/...`, which I think is fine? since other devpod functionality could use this so deleting would help "ignore" the files elsewhere

pascalbreuninger: > True, but only from the content folder, i.e. ~/.devpod/context/...

Yep, we still need to make an exception for local docker and local folder because there it's a direct bind mount to the origin folder

bkneis: ah yes your right, my PR description was wrong I had not tested local dir and local docker. Only local dir with kubernetes, which uses the stream RPCs. The deletion here only happens with a git source (since it's in `CloneRepositoryForWorkspace`). I'm not sure how we can support .devpodignore for the local docker and local dir use case. Unless I symlinked or moved the original folder elsewhere, there will be no way around the bind mount mounting ignored files. But maybe this is fine? Since if you are using local docker and dir, you proboably want everything mounted? What do you think? Can you see a way of supporting?

pascalbreuninger: Oh nice, I've totally missed this! Thanks for clarifying

pascalbreuninger (approved):
LGTM

tcapelle:
Also this only works with cloned repos, not local repos...

PR #1340, opened 2024-10-29
Add github workflow to mark issues as stale then delete after 2 months

bkneis:
I was working through some issues and think it would be nice to have some pipeline clear old issues that are not relevant. I copied this from

https://docs.github.com/en/actions/use-cases-and-examples/project-management/closing-inactive-issues

Is 3 months long enough? Any issues labelled as an enhancement or assign / in progress will be ignored so we can keep long lived feature requests for the road map. I think if a bug goes 3 months and either no one can reproduce or the reporter does not respond closing it is safe.

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -1,0 +1,22 @@
+name: Close inactive issues
+on:
+  schedule:
+    - cron: "30 1 * * *"
+
+jobs:
+  close-issues:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


pascalbreuninger:
Great idea 🙌 
3 months for non-enhancements should be fine, chances that someone comes back are quite low after that time :D And if they do we can still reopen it

pascalbreuninger (approved):
LGTM

Issue #1185, opened 2024-07-23
Support for Zed


brandondrew:
It would be very helpful if DevPod supported the Zed editor.

It's growing in popularity and now available for all 3 major platforms.

pascalbreuninger:
Hi @brandondrew, thanks for opening the issue. 
Zed definitely looks like a welcome addition to the IDE ecosystem and I'm personally excited to see their development. It'd be fantastic if we could add them as an IDE to DevPod. We're currently lacking a way of programatically opening a remote environment though, I've created a [ticket](https://github.com/zed-industries/zed/issues/15070) over at Zeds repo for this purpose.

You can still use Zed with DevPod manually if you want to:
1. Create your workspace with DevPod
2. In Zed (Preview), start the `projects: open remote` action 
3. Click on `New Server`
4. Select `Connect via SSH (default)` and enter either `ssh $WORKSPACE_ID.devpod` or `devpod ssh $WORKSPACE_ID` as the host
5. Zed should install itself into the container and connect your local instance to it

brandondrew:
@pascalbreuninger Thanks!

simlal:
Hey man, I've been trying first with the example from the vscode repo for a quick check. I get this 
![image](https://github.com/user-attachments/assets/45818ce2-f9e7-4a0c-91e3-868a75d81345)

Even though I can ssh into it:
```bash
vscode-remote-try-python on  main via  v3.11.2 on   (ca-central-1)
❯ ssh vscode-remote-try-python.devpod
vscode ➜ /workspaces/vscode-remote-try-python (main) $ cd .local/
vscode ➜ /workspaces/vscode-remote-try-python/.local (main) $ ls -a
.  ..
vscode ➜ /workspaces/vscode-remote-try-python/.local (main) $ which gzip
/usr/bin/gzip

# Checking zed bin and glibc version
vscode ➜ ~/.local/bin $ ls
flask
vscode ➜ ~/.local/bin $ ldd --version
ldd (Debian GLIBC 2.36-9+deb12u8) 2.36
Copyright (C) 2022 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
Written by Roland McGrath and Ulrich Drepper.
```

I am running 
❯ zed --version
Zed  preview 0.157.2

[...]

PR #1341, opened 2024-10-30
Implement experimental support for zed

bkneis:
Fixes https://github.com/loft-sh/devpod/issues/1185

This PR adds experimental support for zed.  Please note, according to [the docs](https://zed.dev/docs/remote-development#known-limitations) remote development in zed does not support extensions. So I have not added an "Install" function to download and run the zed remote server on the container like we do vscode or jetbrains

### DIFF
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -330,6 +330,9 @@ func (cmd *UpCmd) Run(
 				cmd.GitUsername,
 				cmd.GitToken,
 				log)
+
+		case string(config.IDEZed):
+			return zed.Open(ctx, ideConfig.Options, config2.GetRemoteUser(result), result.SubstitutionContext.ContainerWorkspaceFolder, client.Workspace(), log)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


ConradIrwin:
@bkneis if you can wait a week or two, I've just added support for `zed://ssh/` - will ship in 0.160 (likely about 7 days from now).

Then you'll be able to just `open zed://ssh/%s.devpod/%s` instead of needing to try and find the path to the binary.

pascalbreuninger:
@ConradIrwin nice! The change seems to have gone in after `v0.160.0-pre`. If you give us another prerelease we can prepare the change already :)

bkneis:
@ConradIrwin awesome thank you very much! Sure I will wait then update my PR

ConradIrwin:
Should be out in v0160.1, though this is still a preview release. It should hit stable on Wednesday!

Thanks for adding Zed support here :D

PR #1343, opened 2024-10-31
Implement experimental support for marimo

bkneis:
Add experimental support for marimo.

Tested using `./devpod-cli --ide marimo up https://github.com/microsoft/vscode-remote-try-python`

### DIFF
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -5,12 +5,14 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"math/rand"
 	"net"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/blang/semver"
 	"github.com/loft-sh/devpod/cmd/flags"
@@ -27,6 +29,7 @@ import (
 	"github.com/loft-sh/devpod/pkg/ide/fleet"
 	"github.com/loft-sh/devpod/pkg/ide/jetbrains"
 	"github.com/loft-sh/devpod/pkg/ide/jupyter"
+	"github.com/loft-sh/devpod/pkg/ide/marimo"
 	"github.com/loft-sh/devpod/pkg/ide/openvscode"
 	"github.com/loft-sh/devpod/pkg/ide/vscode"
 	"github.com/loft-sh/devpod/pkg/loft"
@@ -63,6 +66,7 @@ type UpCmd struct {
 
 	DotfilesSource string
 	DotfilesScript string
+	Token          string
 }
 
 // NewUpCmd creates a new up command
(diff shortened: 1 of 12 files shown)


cmd/up.go
```diff
@@ -63,6 +66,7 @@ type UpCmd struct {
 
 	DotfilesSource string
 	DotfilesScript string
+	Token          string
```
pascalbreuninger, line 69: What do you think of putting this token into marimo specific settings? Similar to the `bindaddressoption`. 
This way we don't have to pollute all commands with it and can read the ide configuration on the agent through `workspaceInfo.IDE`

bkneis: 100%

bkneis: actually I can't do that unless we are happy with the access token not changing? Currently we generate a new token every connection and exchange over a secure connection. Using the settings would require that the token not change :thinking:

pkg/ide/marimo/marimo.go
```diff
@@ -0,0 +1,104 @@
...
+	out, err := exec.Command(args[0], args[1:]...).CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("error installing jupyter notebook: %w", command.WrapCommandError(out, err))
+	}
+	return s.start()
+}
+
+func (s *Server) start() error {
+	return single.Single("marimo.pid", func() (*exec.Cmd, error) {
+		s.log.Infof("Starting jupyter notebook in background...")
```
pascalbreuninger: 🤭

pkg/ide/marimo/marimo.go
```diff
@@ -0,0 +1,104 @@
...
+		args = append(args, "su", s.userName, "-c", runCommand)
+	} else {
+		args = append(args, "sh", "-c", runCommand)
+	}
+
+	// install
+	s.log.Infof("Installing marimo...")
+	out, err := exec.Command(args[0], args[1:]...).CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("error installing jupyter notebook: %w", command.WrapCommandError(out, err))
```
pascalbreuninger: here as well

pkg/ide/ideparse/parse.go
```diff
@@ -143,6 +143,13 @@ var AllowedIDEs = []AllowedIDE{
 		Icon:         "https://devpod.sh/assets/positron.svg",
 		Experimental: true,
 	},
+	{
+		Name:         config.IDEMarimo,
+		DisplayName:  "Marimo",
+		Options:      vscode.Options,
+		Icon:         "https://devpod.sh/assets/positron.svg",
```
pascalbreuninger: need an icon

pascalbreuninger: nevermind :D

cmd/up.go
pascalbreuninger: Is that a problem for any given session? 
You could stop it, change the token and restart the workspace 🤷‍♂️

bkneis: True, updated :)

pascalbreuninger (approved):
LGTM

pascalbreuninger (approved):
LGTM

PR #1345, opened 2024-10-31
Add experimental support for jupyter desktop

bkneis:
Tested using ./devpod-cli --ide jupyterdesktop --id j5 up https://github.com/microsoft/vscode-remote-try-python

Since we already have the jupyter server logic we only needed to find and enable the client :)


pascalbreuninger (approved):
LGTM

Issue #1198, opened 2024-08-01
VSCodium support is missing, while it exists for VSCode and VSCode for Insiders


KhazAkar:
**Which solution do you suggest?**  
Similar as for VSCode for Insiders or VSCode :)

**Which alternative solutions exist?**  
Remote via SSH or just launching it with devpod plugin while vscodium is open.

**Additional context**  
VSCodium is a project which provides an open source binaries for VSCode

bkneis:
Hi @KhazAkar thanks for the suggestion, I was not aware of vs codium existed! It looks like an interesting project. One solution to this problem could be to symlink vs codium binary as `vscode` in your bin. Depending on your set up something like `ln -s /usr/bin/codium /usr/bin/code`

3timeslazy:
@KhazAkar the problem with VS Codium is that it doesn’t support dev containers. VS Code DevContainer extension is proprietary and won’t work with VS Codium even if you download vsix package from the marketplace. See: https://github.com/VSCodium/vscodium/issues/1229.

The best you can get is to create a container using devpod cli and connect to it via https://github.com/jeanp413/open-remote-ssh

Alternatively, you can try this: https://github.com/3timeslazy/vscodium-devpodcontainers

I faced the same problem some time ago and made this as a workaround. It does exactly what I’ve described: runs devpod up and connects to it using ssh extension from above. It works on two my machines, but very raw and only used by me.

3timeslazy:
@KhazAkar wow, I didn’t know that. Glad to hear this! Will be more careful with new releases now hah

In fact, DevPod supporting Codium would be very cool and beneficial to both the Codium and DevPod communities for at least two reasons:
1.  **Supporting a more private and truly open VS Code alternative**. DevPod users might see Codium on the list of supported IDEs and get interested in it.
2. **Promoting DevPod to Codium users**. Codium users are currently excluded from the devcontaineirs movement due to a lack of open implementations. DevPod integration could encourage devs to develop open alternatives for remote development extensions such as dev containers and remote k8s development. All of that will be good promotion for DevPod among Codium users.

KhazAkar:
@bkneis I will check it tomorrow, but it might work. Currently plugin by 3timeslazy opens it in VSCodium, but having CLI "devpod up" do it automagically, even by linking, will be awesome :)

bkneis:
[...]

PR #1348, opened 2024-11-01
Implement experimental support for codium

bkneis:
Fixes https://github.com/loft-sh/devpod/issues/1198 by using the opsn SSH extension available in the codium marketplace

### DIFF
--- a/pkg/ide/vscode/vscode.go
+++ b/pkg/ide/vscode/vscode.go
@@ -282,22 +282,22 @@ func (o *VsCodeServer) findServerBinaryPath(location string) string {
 	}
 
 	if o.flavor == FlavorCodium {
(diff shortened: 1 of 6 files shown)


pkg/ide/vscode/vscode.go
```diff
@@ -280,6 +281,48 @@ func (o *VsCodeServer) findServerBinaryPath(location string) string {
 		return binPath
 	}
 
+	if o.flavor == FlavorCodium {
+		// check legacy location `$HOME/.vscode-server/bin`
```
3timeslazy, line 285: I think it should be .vscodium-server (at least that's what I have)

pascalbreuninger (approved):
LGTM

PR #1377, opened 2024-11-13
Use sleepmode.loft.sh/last-activity instead of sleep mode config when disabled

bkneis:


pkg/workspace/list.go
```diff
@@ -162,10 +163,18 @@ func listProWorkspaces(ctx context.Context, devPodConfig *config.Config, log log
...
 
 			// last used timestamp
-			lastUsedTimestamp := types.Time{}
+			var lastUsedTimestamp types.Time
 			sleepModeConfig := instance.Status.SleepModeConfig
 			if sleepModeConfig != nil {
 				lastUsedTimestamp = types.Unix(sleepModeConfig.Status.LastActivity, 0)
+			} else {
+				var ts int64
+				if val, ok := instance.Annotations["sleepmode.loft.sh/last-activity"]; ok {
```
pascalbreuninger, line 172: `instance.annotations` could be nil

pascalbreuninger: other than that LGTM

bkneis: good spot :detective:

pascalbreuninger (approved):
LGTM

PR #1388, opened 2024-11-14
Adds ability for user to define additional tags for images

fcaroline2020:
Adds new build cli flag `tag` that will tag the image with the user defined tag and then push

Example CLI command
`./devpod-cli build . --devcontainer-path ./devcontainer.json --force-build --debug --tag test,arm64`


pascalbreuninger:
@fcaroline2020 Looks like there are some linting issues in the PR, once they are updated we're good to merge imo

PR #1390, opened 2024-11-16
improve clarity on provider options and set_option

abdulari:
improve CLI clarity for `provider options` and `provider set-options` that both will accept optional `provider` parameter.


pascalbreuninger (approved):
LGTM, thank you for the contribution @abdulari 🙌

PR #1411, opened 2024-11-23
Revert "fix(pro): wait for outer tunnels to exit and clean up properly"

pascalbreuninger:
Reverts loft-sh/devpod#1396


pascalbreuninger:
@bkneis I want to revert this and try another route - it looks like waiting for the goroutines to finish is technically correct but does surface race conditions that surface every couple of tries. Reverting doesn't fix the race conditions but figuring that out should be part of a overall cleanup in less hectic times

PR #1413, opened 2024-11-25
Add CombinedLogger that can handle multiple Logger instances

janekbaraniewski:


.github/workflows/e2e-tests.yaml
```diff
@@ -117,6 +117,14 @@ jobs:
         with:
           go-version: 1.21.8
 
+      - name: Add repository to Git safe directory list
+        run: git config --global --add safe.directory C:/actions-runner/_work/devpod/devpod
```
pascalbreuninger, line 121: think we don't need this anymore, right?

pascalbreuninger (approved):
Need to remove the e2e changes, other than that LGTM

.github/workflows/e2e-tests.yaml
janekbaraniewski: ah yeah, I've started debugging it in this PR. cleaning up 👍

PR #1423, opened 2024-11-26
POD-1081 | Add option to use specific version od DevPodEnvironmentTemplate

janekbaraniewski:
Fixes POD-1081


pascalbreuninger (approved):
LGTM

PR #1426, opened 2024-11-27
fix: Ensure that the pro provider process is terminated before discon…

PRTTMPRPHT:
…necting it

### DIFF
--- a/desktop/src/contexts/DevPodContext/Pro/ProProvider.tsx
+++ b/desktop/src/contexts/DevPodContext/Pro/ProProvider.tsx
@@ -1,6 +1,6 @@
 import { ProClient, client as globalClient } from "@/client"
 import { ToolbarActions, ToolbarTitle } from "@/components"
-import { Annotations } from "@/lib"
+import { Annotations, Result } from "@/lib"
 import { Routes } from "@/routes"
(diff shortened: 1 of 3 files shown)


desktop/src/contexts/DevPodContext/Pro/ProProvider.tsx
```diff
@@ -77,11 +83,20 @@ export function ProProvider({ host, children }: { host: string; children: ReactN
...
+      return toCancel().finally(() => setWaitingForCancel(false))
+    }
+
+    setCancelWatch({ fn: canceler })
+
+    return () => {
+      canceler()
     }
-  }, [client, store, currentProject])
+  }, [client, store, currentProject, setCancelWatch, setWaitingForCancel])
```
pascalbreuninger, line 99: you shouldn't need these here, they are guaranteed to never change

pascalbreuninger (approved):
small nit, other than that LGTM

desktop/src/contexts/DevPodContext/Pro/ProProvider.tsx
PRTTMPRPHT: Seems we have a different eslint config in this project. Usually, if they are missing, this is flagged.
Changed it.

PR #1436, opened 2024-11-29
feat: Add search function to action logs

PRTTMPRPHT:


desktop/src/views/Pro/Workspace/Logs.tsx
```diff
@@ -144,18 +149,161 @@ type TActionTerminalProps = Readonly<{
...
+  const [debouncedSearchString, setDebouncedSearchString] = useState<string | undefined>(undefined)
+  const [caseSensitive, setCaseSensitive] = useState<boolean>(false)
+  const [wholeWordSearch, setWholeWordSearch] = useState<boolean>(false)
+
+  const searchInputRef = useRef<HTMLInputElement | null>(null)
+
+  // Debounce to prevent stutter when having a huge amount of results.
+  useEffect(() => {
+    // Sneaky heuristic:
+    // If we have more than two characters, we're likely to have a more sane amount of results, so we can skip debouncing.
```
pascalbreuninger, line 163: :D

desktop/src/views/Pro/Workspace/Logs.tsx
```diff
@@ -144,18 +149,161 @@ type TActionTerminalProps = Readonly<{
 }>
 function ActionTerminal({ actionID }: TActionTerminalProps) {
   const action = useAction(actionID)
-  const { terminal, connectStream, clear: clearTerminal } = useStreamingTerminal()
+
+  const [searchString, setSearchString] = useState<string | undefined>(undefined)
```
pascalbreuninger, line 153: Could you wrap these controls into a separate hook/component and also add them to the open source logs please? 
`views/Actions/Action.tsx`

desktop/src/views/Pro/Workspace/Logs.tsx
```diff
@@ -23,10 +27,11 @@ import {
   VStack,
 } from "@chakra-ui/react"
 import dayjs from "dayjs"
-import { useEffect, useState } from "react"
+import { JSXElementConstructor, ReactElement, useEffect, useMemo, useRef, useState } from "react"
```
pascalbreuninger, line 30: couple of leftover imports it seems like?

desktop/src/components/Terminal/useTerminalSearch.tsx
```diff
@@ -0,0 +1,402 @@
...
+    debounce: undefined,
+    activeSearchResult: 0,
+    preWrappedLines: undefined,
+    searchResults: [],
+    highlights: [],
+  })
+
+  const clearDisposables = useCallback(() => {
+    const toClear = searchStateRef.current.disposables
+    searchStateRef.current.disposables = []
```
pascalbreuninger, line 73: Could you swap this line with 74 please? This works because we're assigning a new array to `disposables` but I tripped over it when reading and had to think for a moment

desktop/src/components/Terminal/Terminal.tsx
```diff
@@ -124,6 +171,36 @@ export const Terminal = forwardRef<TTerminalRef, TTerminalProps>(function T({ fo
           terminalRef.current?.writeln(data)
           termFitRef.current?.fit()
         },
+        highlight(row: number, col: number, len: number, color: string, invertText: boolean) {
```
pascalbreuninger, line 174: super nit: renaming to `startCol` would make it a bit clearer that we are highlighting by `startColumn + length`. 
Feel free to ignore

desktop/src/components/Terminal/useTerminalSearch.tsx
```diff
@@ -0,0 +1,402 @@
...
+      }
+    },
+    [searchStateRef, terminalRef]
+  )
+
+  // When the terminal is resized, we need to re-calculate the highlights and jump anchors,
+  // as these have to be positioned per wrapped line.
+  const onResize = useCallback(
+    (cols: number) => {
+      const displayLines = (searchStateRef.current.preWrappedLines = wrapLines(
```
pascalbreuninger, line 112: nit: could you extract this into two lines please?

desktop/src/components/Terminal/useStreamingTerminal.tsx
```diff
@@ -1,47 +1,96 @@
...
           if (event.data.message === undefined) {
             return
           }
-          terminalRef.current?.writeln(formatLine(event.data))
+
+          const formattedLine = formatLine(event.data)
+          terminalRef.current?.writeln(formattedLine.ansi)
+
+          searchStateRef.current.preWrappedLines = undefined
+          searchStateRef.current.searchableLines.push(...processInputLine(formattedLine.plain))
```
pascalbreuninger, line 47: There isn't a good way to reuse the terminals internal state I suppose? 
Is it because of formatting and line breaks/wraps?

PRTTMPRPHT: It's not a very fun API to work with. At this stage you will get the wrong length for the underlying internal buffer (the actual write is delayed via setTimeout internally...) and cannot really access the data properly. I wouldn't rule it out that there is a way to somehow get notified when the write occurs, but at this point I found it much saner to just wrap the lines ourselves rather than trying to build around the terminal's way of doing things.

If we keep running into desync issues (search results appear where they shouldn't) we can revisit this, but right now it works on all the cases that I'm able to test.

desktop/src/views/Pro/Workspace/Logs.tsx
PRTTMPRPHT: Done

desktop/src/components/Terminal/useTerminalSearch.tsx
PRTTMPRPHT: Done

desktop/src/components/Terminal/Terminal.tsx
PRTTMPRPHT: Done

desktop/src/components/Terminal/useTerminalSearch.tsx
PRTTMPRPHT: Done

desktop/src/views/Pro/Workspace/Logs.tsx
PRTTMPRPHT: Done

PR #1442, opened 2024-12-03
POD-1124 | Restart failed credentials server

janekbaraniewski:
Fixes POD-1124

### DIFF
--- a/cmd/ssh.go
+++ b/cmd/ssh.go
@@ -5,6 +5,7 @@ import (
 	"encoding/base64"
 	"fmt"
 	"io"
+	"math"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -30,6 +31,8 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"golang.org/x/crypto/ssh"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


cmd/ssh.go
```diff
@@ -647,8 +647,19 @@ func startWorkspaceCredentialServer(ctx context.Context, client *ssh.Client, use
...
 
-	if err := devssh.Run(ctx, client, command, stdin, stdout, writer, nil); err != nil {
-		return fmt.Errorf("run credentials server: %w", err)
+	for {
+		select {
+		case <-ctx.Done():
+			log.Infof("Context canceled, stopping credentials server: %v", ctx.Err())
+			return nil
+		default:
+			if err := devssh.Run(ctx, client, command, stdin, stdout, writer, nil); err != nil {
```
bkneis, line 656: I think we need a sleep in case of repeated failures

pkg/tunnel/services.go
```diff
@@ -131,10 +132,22 @@ func RunInContainer(
...
-		err = devssh.Run(cancelCtx, containerClient, command, stdinReader, stdoutWriter, writer, nil)
-		if err != nil {
-			return err
+		for {
+			select {
+			case <-cancelCtx.Done():
+				log.Infof("Context canceled, stopping credentials server: %v", cancelCtx.Err())
+				return nil
+			default:
+				err := devssh.Run(cancelCtx, containerClient, command, stdinReader, stdoutWriter, writer, nil)
```
bkneis, line 141: same here

cmd/ssh.go
janekbaraniewski: ah a back off mechanism, yeah good point I think we even talked about it. 👀

pascalbreuninger: You could use `retry.OnError` from `"k8s.io/client-go/util/retry"` if you want something out of the box

pascalbreuninger (approved):
LGTM

PR #1443, opened 2024-12-03
feat: Add selection of presets for DevPod Pro workspace creation

PRTTMPRPHT:


desktop/src/views/Pro/SelectPreset/SelectPreset.tsx
```diff
@@ -0,0 +1,196 @@
...
+        return true
+      }
+
+      if (preset.spec?.source.image && preset.spec.source.image.includes(searchString)) {
+        return true
+      }
+
+      return preset.spec?.source.git && preset.spec.source.git.includes(searchString)
+    })
+  }, [templates?.presets, searchString])
```
pascalbreuninger, line 52: nice

desktop/src/views/Pro/SelectPreset/SelectPreset.tsx
```diff
@@ -0,0 +1,196 @@
...
+              boxSizing={"border-box"}
+              borderRadius={"4px"}>
+              <Button variant={"outline"} colorScheme={"primary"}>
+                New Custom Workspace
+              </Button>
+            </Box>
+          )}
+          {filteredPresets.map((preset, index) => (
+            <PresetBox
+              key={index}
```
pascalbreuninger, line 126: super nit: `preset.metadata.name!` (it's guaranteed to be set by k8s)

desktop/src/views/Pro/CreateWorkspace/CreateWorkspace.tsx
```diff
@@ -34,6 +36,31 @@ export function CreateWorkspace() {
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
+
+  const preset = useMemo(() => {
+    if (!presetId) {
+      return undefined
+    }
+
+    return (presets ?? []).find((p) => p.metadata?.name === presetId)
+  }, [presetId, presets])
```
pascalbreuninger, line 62: Couldn't all of this just go into the `useMemo`?

desktop/src/views/Pro/CreateWorkspace/PresetInput.tsx
```diff
@@ -0,0 +1,81 @@
...
+                  {"No preset"}
+                </MenuItem>
+                {presets?.map((p, i) => (
+                  <MenuItem
+                    fontSize={"md"}
+                    key={i}
+                    onClick={() => {
+                      setPreset?.(p.metadata?.name)
+                    }}>
+                    {p.spec?.displayName ?? p.metadata?.name ?? ""}
```
pascalbreuninger, line 71: nit: `getDisplayName` in case we ever change/add a field

desktop/src/views/Pro/CreateWorkspace/UpdateWorkspace.tsx
```diff
@@ -26,10 +27,28 @@ export function UpdateWorkspace({ instance, template }: TUpdateWorkspaceProps) {
...
 
+  const { data: templates, isLoading: isTemplatesLoading } = useTemplates()
+
+  const presets = templates?.presets
+
+  const [presetId, setPresetId] = useState<string | undefined>(instance.spec?.presetRef?.name)
+
+  useEffect(() => {
+    setPresetId(instance.spec?.presetRef?.name)
+  }, [instance.spec?.presetRef?.name])
```
pascalbreuninger, line 38: also could go into the `useMemo`?

desktop/src/views/Pro/CreateWorkspace/CreateWorkspace.tsx
PRTTMPRPHT: No, I want the router query only to be evaluated on the first render, it will not get updated when I change the preset while in the view. The details of which preset we have, however, can change.

desktop/src/views/Pro/CreateWorkspace/UpdateWorkspace.tsx
PRTTMPRPHT: Arguably you could find a way to merge this one with its following `useMemo` but I'd question why you would want that.

As it is, this effect serves one single purpose: Synchronize the internal state with any new changes pushed from the parent component. Neat and tidy.

The following `useMemo` resolves the internal preset ID - which can also change from other sources(!) - to the full preset data, without any side effects.

This are distinct purposes, I think we should keep it like it is.

desktop/src/views/Pro/SelectPreset/SelectPreset.tsx
PRTTMPRPHT: Done

desktop/src/views/Pro/CreateWorkspace/PresetInput.tsx
PRTTMPRPHT: Done

pascalbreuninger (approved):
LGTM

PR #1452, opened 2024-12-06
feat: Change update server to a custom mirror so we can avoid rate li…

PRTTMPRPHT:
…mits on the clients


PRTTMPRPHT:
@pascalbreuninger This is the PR for the update server change, feel free to merge once the service is deployed

pascalbreuninger (approved):
LGTM overall, small risk that we don't get enough releases since GitHubs default limits are set to 30 releases per page. Since this is only the fallback not too big of an issue though

Issue #1307, opened 2024-10-07
SSH ForwardAgent does not work for openvscode


nrontsis:
**What happened?**  
ForwardAgent does not work on openvscode. If I do e.g.
```
devpod up https://github.com/loft-sh/devpod-example-go --provider docker --ide openvscode
```
the ssh_auth_sock is pointing to a non-existing file.
```
vscode ➜ /workspaces/devpod-example-go (main) $ ssh -T git@github.com
git@github.com: Permission denied (publickey).
vscode ➜ /workspaces/devpod-example-go (main) $ ls $SSH_AUTH_SOCK
ls: cannot access '/tmp/auth-agent3979699691/listener.sock': No such file or directory
```
while manually ssh'ing into the pod via `ssh devpod-example-go.devpod` has no issue: i.e. ForwardAgent works fine, and I can authenticate e.g. in github fine.

**What did you expect to happen instead?**  
ForwardAgent to work for openvscode, as it does when I ssh into the pod.

**How can we reproduce the bug?** (as minimally and precisely as possible)  
Run:
```
devpod up https://github.com/loft-sh/devpod-example-go --provider docker --ide openvscode
```

**Local Environment:**  
- DevPod Version: v0.5.21
- Operating System: linux + mac
- ARCH of the OS: AMD64 + ARM64

**DevPod Provider:**  
- Local/remote provider: docker 20.10.21

**Anything else we need to know?**  
Nothing

bkneis:
@nrontsis thanks for raising this issue and providing a clear summary and example! I have replicated the issue and can confirm the problematic behaviours. After debugging I believe the issue is here https://github.com/loft-sh/devpod/blob/main/pkg/ssh/server/ssh.go#L136. We are overriding the SSH_AUTH_SOCK env var on each handler in the SSH server. I believe the issue is caused due to additional SSH commands that are sent in order to set up the container tunnel via the browser. For instance if you perform `ls -la /tmp/auth-agent*` you can see the correct `listener.sock` just in the wrong temp directory. If you manually update `SSH_AUTH_SOCK` to point to this file it works again.

I think the fix is to change the SSH server implementation so we can specify on a per connection basis if the SSH_AUTH_SOCK should be updated. This might be a fairly big change and I am currently occupied with another feature but will return to this soon!

nrontsis:
thanks a lot for the reply and the analysis @bkneis!

bkneis:
https://github.com/loft-sh/devpod/issues/1449

Issue #1464, opened 2024-12-09
The /tmp directory gets flooded by "auth-agent*" folders


fikovnik:
**What happened?**  

I use devpod to create a devcontainer that I connect from both Intellij IDEA (for the Java part of the project) and ssh where I start tmux and neovim (for the C++ part of the project). Today I realized that the `/tmp` directory gets flooded by `auth-agent*` directories. A new directory is created about every second:

```
drwx------ 1 vscode vscode        0 Dec  9 10:34 auth-agent1008629510
drwx------ 1 vscode vscode        0 Dec  9 10:37 auth-agent1027115269
drwx------ 1 vscode vscode        0 Dec  9 10:36 auth-agent1040715800
drwx------ 1 vscode vscode        0 Dec  9 10:32 auth-agent107829153
drwx------ 1 vscode vscode        0 Dec  9 10:34 auth-agent1096869530
drwx------ 1 vscode vscode        0 Dec  9 10:34 auth-agent1100131597
drwx------ 1 vscode vscode        0 Dec  9 10:34 auth-agent1104985039
drwx------ 1 vscode vscode        0 Dec  9 10:34 auth-agent1147654851
drwx------ 1 vscode vscode        0 Dec  9 10:33 auth-agent117922636
drwx------ 1 vscode vscode        0 Dec  9 10:31 auth-agent1198594084
drwx------ 1 vscode vscode        0 Dec  9 10:31 auth-agent1202486330
```

The main issue that my app does generate some temp files in `/tmp` and because there are quickly too many `auth-agent*` directories, all file system operations become too slow.

**How can we reproduce the bug?** (as minimally and precisely as possible)  

My *`devcontainer.json`*:
```
 "name": "r-compile-server",
  "build": {
    "dockerfile": "./Dockerfile",
    "context": "."
  },
  "workspaceMount": "source=${localWorkspaceFolder},target=/workspace,type=bind,consistency=cached",
  "workspaceFolder": "/workspace",
  "postCreateCommand": "make setup",
  "customizations": {
    "vscode": {
      "extensions": [
        "vscjava.vscode-java-pack",
        "llvm-vs-code-extensions.vscode-clangd",
        "vadimcn.vscode-lldb"
      ]
    }
  },
  "runArgs": [
    "--cap-add=SYS_PTRACE",
    "--security-opt",
    "seccomp=unconfined"
  ],
  "updateRemoteUserUID": true,
  "features": {
[...]

Issue #1456, opened 2024-12-07
$SSH_AUTH_SOCK - No such file or directory 


jleni:
**What happened?**  

Local SSH credentials are not available in the remote host.

Also it seems the socket is not linked/mount correctly?

```
ls -la $SSH_AUTH_SOCK 
ls: cannot access '/tmp/auth-agent2010668863/listener.sock': No such file or directory
```

**What did you expect to happen instead?**  

My *`devcontainer.json`*:
```
{
    "name": "Python Development",
    "image": "mcr.microsoft.com/devcontainers/base:ubuntu",
    "customizations": {
        "vscode": {
            "extensions": [
                "ms-python.python",
                "ms-python.vscode-pylance",
                "anysphere.pyright",
                "redhat.vscode-yaml",
                "ms-python.black-formatter"
            ],
            "settings": {
                "[python]": {
                    "editor.formatOnSave": true
                }
            }
        }
    },
    "postCreateCommand": "sudo apt-get update && sudo apt-get install -y python3 python3-pip git gh"
}
```

**Local Environment:**  
- DevPod Version: 0.6.4
- Operating System: mac
- ARCH of the OS: ARM64 

**DevPod Provider:**  
- Cloud Provider: selfhosted
- Kubernetes Provider: Client Version: v1.31.3 / Kustomize Version: v5.4.2 / Server Version: v1.30.6+rke2r1

**Anything else we need to know?**  

- I am using 1password agent, similarly to https://github.com/loft-sh/devpod/issues/1438
- I am using Cursor as an IDE. (I think this is relevant as when running `devpod ssh` directly from the terminal, the agent DOES work). It fails when using cursor

jleni:
I think this part of the log (when using cursor) is relevant

```
[10:14:41] debug Received ping from agent
[10:14:41] debug Start setting up container...
[10:14:41] debug Credentials server started on port 14951...
[10:14:42] debug Incoming client connection at /
[10:14:42] debug Credentials server started...
[10:14:42] debug Patch etc environment & profile...
[10:14:42] debug Run lifecycle hooks commands...
[10:14:42] debug running user env probe with shell "bash", probe "loginInteractiveShell", user "vscode" and command "cat /proc/self/environ"
[10:14:42] debug Done setting up environment
[10:14:42] debug Setup cursor...
[...]

PR #1471, opened 2024-12-12
Add --reuse-sock flag so browser IDEs can reuse another SSH connections SSH_AUTH_SOCK

bkneis:
This PR introduces the hidden flag `--reuse-sock` to the SSH and helper ssh-server commands so that browser IDE's can first set up a SSH connection and allow the browser tunnel to reuse this.

Fixes https://github.com/loft-sh/devpod/issues/1307, https://github.com/loft-sh/devpod/issues/1456 and https://github.com/loft-sh/devpod/issues/1464

Why do we need this? "normal" IDE's like vscode use an SSH connection to connect to the remote, this connection requersts agent forwarding so the user's SSH_AUTH_SOCK is available in the workspace.

For browser IDEs this SSH connection is not made, the only long lived connection we have is used for the credential server and runs as root, which it must.

Therefore this fix is to emulate the behaviour of the normal IDE and allow the vscode session to highjack the SSH_AUTH_SOCK of the unused "backhaul" connection.

This PR was tested using local docker provider and kubernetes with openvscode and vscode. I tried pro with vscode and it works but using openvscode does not because the devpod CLI on the platform does not have the build changes.

Additionally we perform the clean up of the tmp dir if used


cmd/ssh.go
```diff
@@ -430,6 +433,10 @@ func (cmd *SSHCmd) startTunnel(ctx context.Context, devPodConfig *config.Config,
 
 	log.Debugf("Run outer container tunnel")
 	command := fmt.Sprintf("'%s' helper ssh-server --track-activity --stdio --workdir '%s'", agent.ContainerDevPodHelperLocation, workdir)
+	if cmd.ReuseSSHAuthSock != "" {
+		log.Info("Reusing SSH_AUTH_SOCK")
```
pascalbreuninger, line 437: nit: This will always be printing when we connect, can you make this log level `Debug` please?

cmd/ssh.go
```diff
@@ -484,7 +491,7 @@ func (cmd *SSHCmd) startServices(
 	log log.Logger,
 ) {
 	if cmd.User != "" {
-		err := tunnel.RunInContainer(
+		err := tunnel.RunServices(
```
pascalbreuninger, line 494: Could you revert this naming change please? Makes it harder to backmerge and it's not really required in this PR. 
If you absolutely want to change it, please open a new one

cmd/up.go
```diff
@@ -151,6 +153,19 @@ func (cmd *UpCmd) Run(
...
+	targetIDE := client.WorkspaceConfig().IDE.Name
+	// Check override
+	if cmd.IDE != "" {
+		targetIDE = cmd.IDE
+	}
+	if !cmd.Proxy && ide.ReusesAuthSock(targetIDE) {
+		cmd.SSHAuthSockID = util.RandStringBytes(10)
+		log.Debug("Reusing SSH_AUTH_SOCK", cmd.SSHAuthSockID)
+	} else if cmd.Proxy && ide.ReusesAuthSock(targetIDE) {
+		log.Info("Reusing SSH_AUTH_SOCK is not supported with proxy mode, consider launching the IDE from the platform UI")
```
pascalbreuninger, line 166: This can potentially be confusing if you don't know what `SSH_AUTH_SOCK` is and how it relates to the ssh agent. Could you make this a debug log please

cmd/up.go
```diff
@@ -994,6 +1077,8 @@ func createSSHCommand(
 	}
 	args = append(args, extraArgs...)
 
+	logger.Debug("Connecting with SSH command ", execPath, args)
```
pascalbreuninger, line 1080: Do we still need this?

pkg/tunnel/container.go
```diff
@@ -174,7 +183,8 @@ func (c *ContainerHandler) updateConfig(ctx context.Context, sshClient *ssh.Clie
 	}
 }
 
-func (c *ContainerHandler) runRunInContainer(ctx context.Context, sshClient *ssh.Client, runInContainer Handler, envVars map[string]string) error {
+// runInContainer uses the connected SSH client to execute runInContainer on the remote
+func (c *ContainerTunnel) runInContainer(ctx context.Context, sshClient *ssh.Client, handler Handler, envVars map[string]string) error {
```
pascalbreuninger, line 187: Let's add the doc string but please revert the naming changes 😅

cmd/ssh.go
bkneis: The reason I changed this function's name was because it runs a server locally, so the name is misleading IMO

cmd/up.go
bkneis: I thought it would be a useful debug log so we have the flags the SSH connection made

pascalbreuninger: A lot of people have debug logs turned on and that's going to bloat them quite a bit since some of the commands include a compressed version of `workspaceInfo`. I don't think it's worth the noise.
We could change this to the `trace` level maybe?

bkneis: Ah yes good point, I'll remove then and add when I need it :) I think this is where structured logging can really be useful

pascalbreuninger (approved):
LGTM 🙌

PR #1484, opened 2024-12-17
fix: Ensure that parameter default values are correctly set when crea…

PRTTMPRPHT:
…ting a pro instance and switching template versions


pascalbreuninger (approved):
LGTM

Issue #1481, opened 2024-12-17
Build downgrades existing docker syntax specification


agrinh:
**What happened?**  
Trying to build a `Dockerfile` with an existing syntax specification fails if a recent syntax version is used since the syntax is replaced in the generated `Dockerfile-with-features` with `docker.io/docker/dockerfile:1.4`.

**What did you expect to happen instead?**  
devpod should never force an older syntax version, but could potentially bump the version to a minimum one if required.

**How can we reproduce the bug?** (as minimally and precisely as possible)  

Run `devpod build ... --skip-push --debug` with a `Dockerfile` like this:

```Dockerfile
# syntax=docker/dockerfile:1

FROM ubuntu:latest AS dev
ADD --checksum=sha256:b8713c36d3e6f6b2520830180cf6b5779e48410ad5d5e1b192635b0c7359fcc6 \
    https://packages.microsoft.com/config/ubuntu/22.04/packages-microsoft-prod.deb \
    /tmp/packages-microsoft-prod.deb
```

This will fail with `info ERROR: failed to solve: dockerfile parse error on line 2: unknown flag: checksum`. The logs reveal the reason, namely that the syntax is overridden with 1.4: `info #2 resolve image config for docker.io/docker/dockerfile:1.4`

This also happens if using a complete version e.g.  `# syntax=docker/dockerfile:1.12`.

**Local Environment:**  
- DevPod Version: v0.6.5
- Operating System: mac
- ARCH of the OS: ARM64

**DevPod Provider:**  
- Cloud Provider: N/A
- Kubernetes Provider: N/A
- Local/remote provider: docker
- Custom provider: N/A

**Anything else we need to know?**

bkneis:
Hey @agrinh thanks for reporting your issue! I've reproduced this locally and working on and testing a fix now

PR #1493, opened 2024-12-19
Feature/fix builds docker syntax

bkneis:
This PR fixes https://github.com/loft-sh/devpod/issues/1481

In summary devpod always assumes a build syntax of docker/dockerfile:1.4. The reference CLI parses out the build syntax and uses it in the feature layers - https://github.com/devcontainers/cli/blob/main/src/spec-node/containerFeatures.ts#L200

This PR does the same falling back to 1.4 if no build syntax is specified.

I tested this using examples/build and adding the problematic build commands such as `RUN --checksum` and watching it fail then work with the PR

### DIFF
--- a/examples/build/Dockerfile
+++ b/examples/build/Dockerfile
@@ -1,3 +1,4 @@
+# syntax=docker/dockerfile:1
 FROM mcr.microsoft.com/devcontainers/go:1.22-bullseye
 
 ARG TARGETOS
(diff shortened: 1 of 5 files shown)


pascalbreuninger (approved):
LGTM

Edit: Looks like we've hit a nil pointer somewhere in here. Could you pass along the syntax instead of the full struct?

Issue #1448, opened 2024-12-05
Output of xdg-open not visible when something goes wrong


mathroc:
**What happened?**  

I had an issue with my xdg config for Zed that prevented to start Zed from devpod CLI, with just those logs:

> 19:11:18 info Opening Zed editor...
19:11:21 debug Starting Zed caused error: exit status 3
19:11:21 error Seems like you don't have Zed installed on your computer locally

I didn't know what was going on so I opened an issue : https://github.com/loft-sh/devpod/issues/1444 where I was told it was executing the command `xdg-open zed://...`. After that I was able to fix the issue

**What did you expect to happen instead?**  

I expected to have more info in the log about what was going on, I think 2 things could be improved:

* in debug mode, print the command that is going to be executed. I would have been able to reproduce the issue outside of devpod and figure out how to solve it on my own
* bonus: display the output of the xdg-open command or save it somewhere and log a debug message to look there

**How can we reproduce the bug?** (as minimally and precisely as possible)  

On Linux, try starting a workspace with an editor not available on the machine

**Local Environment:**  
- DevPod Version: v0.6.3~
- Operating System: linux
- ARCH of the OS: AMD64 

**DevPod Provider:**  
- Local/remote provider: docker

**Anything else we need to know?**

acolombier:
I am facing a somewhat similar issue. Looks like when installed via Flatpak, `xdg-open ` may returns a sucess exit code, but not do anything.

I have wrote a patch for this

```diff
diff --git a/pkg/ide/zed/zed.go b/pkg/ide/zed/zed.go
index fdf3a78c..c9f249f2 100644
--- a/pkg/ide/zed/zed.go
+++ b/pkg/ide/zed/zed.go
@@ -3,6 +3,7 @@ package zed
 import (
        "context"
        "fmt"
+       "os/exec"
 
        "github.com/loft-sh/log"
        "github.com/skratchdot/open-golang/open"
@@ -10,12 +11,26 @@ import (
        "github.com/loft-sh/devpod/pkg/config"
 )
 
+var FlatpakPackageName string = "dev.zed.Zed"
+
 // Open first finds the zed binary for the local platform and then opens the zed editor with the given workspace folder
 func Open(ctx context.Context, values map[string]config.OptionValue, userName, workspaceFolder, workspaceID string, log log.Logger) error {
        log.Info("Opening Zed editor...")
 
[...]

PR #1494, opened 2024-12-19
Print xdg output for linux

bkneis:
This PR outputs the xdg-open output as debug if the command fails to help linux users debug xdg issues. Fixes https://github.com/loft-sh/devpod/issues/1448


pascalbreuninger (approved):
LGTM

Issue #1438, opened 2024-12-02
Starting container with --debug prints repeating fatal error 


ptab:
**What happened?**
~~I am unable to interact with any git repository inside a devcontainer, as DevPod is not able to forward SSH authentication requests to my SSH agent.~~
Turns out SSH forwarding does works, it was an issue on my side. Somehow MacOS replaced `SSH_AUTH_SOCK` with its own agent, which apparently would still internally forward requests to 1Password (since I would get the 1Password fingerprint dialogs when connecting directly via SSH).
I do still see a lot of `fatal` error messages when connecting to the container using `devpod ssh --debug` so I'll leave this issue up.

**What did you expect to happen instead?**  
~~I expect DevPod to correctly forward requests from inside the container to my SSH agent so that I am able to interact with a git repository.~~
I expect no fatal errors in the debug logs when connecting to a devcontainer.

**How can we reproduce the bug?** (as minimally and precisely as possible)  
I created a simple project to showcase the issue: https://github.com/ptab/reproduce-devpod-bug. 
1. Create a workspace from the repository:
   ```
   ❯ devpod up git@github.com:ptab/reproduce-devpod-bug.git --ide none
   16:47:15 info Creating devcontainer...
   16:47:15 info Clone repository
   16:47:15 info URL: git@github.com:ptab/reproduce-devpod-bug.git
   16:47:17 done Successfully cloned repository
   16:47:17 warn dev container config is missing one of "image", "dockerFile" or "dockerComposeFile" properties, defaulting to auto-detection
   16:47:18 info Found existing local image vsc-content-32bbf:devpod-cd4037f429101ea3b6dd5207e681a1c8
   16:47:18 info Inspecting image vsc-content-32bbf:devpod-cd4037f429101ea3b6dd5207e681a1c8
   16:47:18 info 727e017f3fd721b413f857fd14038d6bccebe09e71844ac59e408025946f377b
   16:47:19 info Setup container...
   16:47:20 info Chown workspace...
   16:47:20 info Run command : mkdir ~/.ssh && ssh-keyscan -t rsa github.com >> ~/.ssh/known_hosts...
   16:47:20 warn # github.com:22 SSH-2.0-babeld-9e2e2a76e
   16:47:20 done Successfully ran command : mkdir ~/.ssh && ssh-keyscan -t rsa github.com >> ~/.ssh/known_hosts
[...]

PR #1496, opened 2024-12-19
Base 64 encode and decode signing key

bkneis:
Fixes https://github.com/loft-sh/devpod/issues/1438

Not sure how to test this one. I ran up using the problematic workspace and could SSH and use my git credentials. But since my signing key doesn't cause an issue I can't replicate. But I can confirm the base64 decode / encode works


pascalbreuninger (approved):
LGTM

Issue #1490, opened 2024-12-19
Devpod fails to pull prebuilt images on ARM64


Jacob-Morgan:
**What happened?**  
Using prebuilt images on ARM64 fails to pull the image from the remote registry resulting in an error, after which the container is built locally.

```
14:12:57 debug Try to find prebuild image devpod-8d33aac0ff64f50b048d143deefe758b in repositories XXXXXXXXX.dkr.ecr.us-east-1.amazonaws.com/devcontainers,XXXXXXXXX.dkr.ecr.us-east-1.amazonaws.com/devcontainers
14:12:58 debug Error trying to find prebuild imageXXXXXXXXX.dkr.ecr.us-east-1.amazonaws.com/devcontainers:devpod-8d33aac0ff64f50b048d143deefe758b: 
  retrieve image XXXXXXXXX.dkr.ecr.us-east-1.amazonaws.com/devcontainers:devpod-8d33aac0ff64f50b048d143deefe758b: 
  no child with platform linux/amd64 in index XXXXXXXXX2.dkr.ecr.us-east-1.amazonaws.com/devcontainers:devpod-8d33aac0ff64f50b048d143deefe758b
```

The reported error is correct `no child with platform linux/amd64` as only the `linux/arm64` platform image exists. Looking at the code the issue is caused by the defaults used by the underlying go package `go-containerregistry`.  When getting an image unless a platform is provided it defaults to `linux/amd64` regardless of actual platform being used.

https://github.com/loft-sh/devpod/blob/7445be138c7a63e953fb4a271047a5acb81192cd/pkg/image/image.go#L32

Patching the code base with, results in the image being pulled correctly and reused for future containers 

```
	remoteOptions := []remote.Option{
		remote.WithAuthFromKeychain(keychain),
		remote.WithPlatform(v1.Platform{Architecture: "arm64", OS: "linux"}),
	}

	img, err := remote.Image(ref, remoteOptions...)
```

[// The default platform is amd64/linux.](https://github.com/google/go-containerregistry/blob/6bce25ecf0297c1aa9072bc665b5cf58d53e1c54/pkg/v1/remote/options.go#L218)

**What did you expect to happen instead?**  
It should automatically detect and set the remote image options based on the local platform, potentially even override the image using an CLI option `--platform`.   

**How can we reproduce the bug?** (as minimally and precisely as possible)  
It happens every time when using an ARM64 platform, the image is built and upload but will not be pulled.

**Local Environment:**  
[...]

PR #1499, opened 2024-12-19
Detect arch when getting image for prebuild

bkneis:
Fixes https://github.com/loft-sh/devpod/issues/1490

I can't test this though since I am not on ARM :(

I believe there should be a follow up task in improving how we get the image from the registry for kubernetes provider. Currently there are other instances of image.GetImage but have no way of knowing the arch yet. Most other instances also only need to know if the image exists which can be done more cost effectively then getting the entire image. Additionally performing this check requires the local user to be authenticated to the registry.


pascalbreuninger (approved):
LGTM 

> Most other instances also only need to know if the image exists which can be done more cost effectively then getting the entire image. 
Agree, we can potentially just use `remote.Get` and see if we get back a valid manifest

Issue #1510, opened 2024-12-24
Permission denied running build with unset TMPDIR


mikedlr:
**What happened?**  

mikedlr@kinga:~/dev/devpod/devpod$ ./hack/rebuild.sh 
[INFO] Building for linux/amd64
[INFO] Building for linux/arm64
[INFO] Building for windows/amd64
[INFO] Building for darwin/amd64
[INFO] Building for darwin/arm64
[INFO] Built binaries for all platforms in test/ directory
[sudo] password for mikedlr: 
[INFO] Built devpod binary and moved to /usr/local/bin
[INFO] Copied binaries to desktop/src-tauri/bin
mkdir: cannot create directory ‘/devpod-cache’: Permission denied

**What did you expect to happen instead?**  

mikedlr@kinga:~/dev/devpod/devpod$ ./hack/rebuild.sh 
[INFO] Building for linux/amd64
[INFO] Building for linux/arm64
[INFO] Building for windows/amd64
[INFO] Building for darwin/amd64
[INFO] Building for darwin/arm64
[INFO] Built binaries for all platforms in test/ directory
[INFO] Built devpod binary and moved to /usr/local/bin
[INFO] Copied binaries to desktop/src-tauri/bin
[INFO] Copied binaries to /tmp/devpod-cache

**How can we reproduce the bug?** (as minimally and precisely as possible)  

    unset TMPDIR
    ./hack/rebuild.sh 

**Local Environment:**  
- DevPod Version:  ./hack/rebuild.sh 
- Operating System:  ubuntu
- ARCH of the OS: AMD64

**DevPod Provider:**  

- Local/remote provider: docker | ssh

**Anything else we need to know?**

PR #1511, opened 2024-12-24
Provide a default value for TMPDIR in the case none is given

mikedlr:
Simple fix - if TMPDIR isn't set, default it to /tmp

resolves **https://github.com/loft-sh/devpod/issues/1510**


bkneis (approved):
LGTM! Thanks for the contribution

PR #1520, opened 2025-01-06
cli: add an environment variable to disable telemetry globally

folliehiyuki:
With the current model described at <https://devpod.sh/docs/other-topics/telemetry#telemetry-opt-out-process>, a CLI user wanting to opt out of telemetry completely needs to remember to set the TELEMTRY option for every new context.

This PR adds an additional environment variable as a method to disable telemetry regardless of the current context. This is a common approach among open-sourced CLI programs.

### DIFF
--- a/docs/pages/other-topics/telemetry.mdx
+++ b/docs/pages/other-topics/telemetry.mdx
@@ -47,7 +47,7 @@ Below you can find an example of the payload that DevPod CLI would send to our t
 
 ### Telemetry opt-out process
 
-To disable the telemetry execute the following command:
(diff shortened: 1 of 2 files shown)


pascalbreuninger:
@folliehiyuki thanks for your contribution.
Fair enough, could you also update the docs please?

folliehiyuki:
I've updated the telemetry docs page. My editor spotted 2 spelling errors in the file, so I added another commit fixing those as well, if you don't mind.

pascalbreuninger (approved):
LGTM

PR #1521, opened 2025-01-06
Replace the usage of go-homedir with os.UserHomeDir()

folliehiyuki:
`go-homedir` project has been archived for a long time. `os.UserHomeDir()` returns the same value set so we can use it as a drop-in replacement. Sadly, some vendored modules still depend on `go-homedir`. The offended modules are `github.com/awslabs/amazon-ecr-credential-helper` and `github.com/Azure/go-autorest`, which are both pulled by the direct dependency `github.com/google/go-containerregistry/pkg/authn/k8schain`.

Ref: https://github.com/mitchellh/go-homedir/issues/34

### DIFF
--- a/cmd/pro/start.go
+++ b/cmd/pro/start.go
@@ -21,7 +21,6 @@ import (
 	"github.com/denisbrodbeck/machineid"
 	jsonpatch "github.com/evanphx/json-patch"
 	"github.com/mgutz/ansi"
-	"github.com/mitchellh/go-homedir"
 	"github.com/skratchdot/open-golang/open"
 
 	storagev1 "github.com/loft-sh/api/v4/pkg/apis/storage/v1"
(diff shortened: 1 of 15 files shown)


pascalbreuninger:
@folliehiyuki thanks for the contribution!
I understand that it's simpler to use `os.UserHomeDir()` instead of `go-homedir` but go-homedir has a couple of nice fallbacks like reading `/etc/passwd` if the user env var isn't specified. I don't exactly know if it will break if we change it but I'd also rather not take our chances here.

Since go-homedir is archived and small enough it might be just fine to put it's code somewhere under `/pkg`, wyt?

folliehiyuki:
Sound okay to me. I've copied the code from `go-homedir` with some modifications into `pkg/util`. If you prefer somewhere else I'll update it.

The only difference I see between `go-homedir` and `os.UserHomeDir()` is that the former one always checks for $HOME even on Windows, so I added a test case for it. I don't use Windows so haven't actually verify the test.

pascalbreuninger:
Sounds good, thank you!

pascalbreuninger (approved):
LGTM 🙌

pascalbreuninger:
The linting CI failed and I've fixed it quickly. Hope you don't mind @folliehiyuki

Issue #1433, opened 2024-11-29
Devpod build fails when using Multi-Stage build with a Feature


joshbrooks:
**What happened?**

In my dev environment I use a multistage build, like this

```dockerfile
FROM mcr.microsoft.com/devcontainers/python:1-3.12-bookworm AS builder
# ...
FROM builder AS final
# ...
```

I also have in my devcontainer.json a features clause to add node
```
	"features": {
		"ghcr.io/devcontainers/features/node:1": {
			"installYarnUsingApt": true,
			"version": "lts",
			"nvmVersion": "latest"
		}
	},
```

When I try to start a workspace (debug on) it fails. this debug log indicates that an incorrect `dockerfile` is generated:
Note line 3 `FROM  AS final`.

```

{"type":"data","data":{"time":"2024-11-29T11:14:54.363901867+09:00","message":"Creating extended Dockerfile /var/home/josh/.devpod/agent/contexts/default/workspaces/education-git-rc-2-0/content/.devcontainer/python/.devpod-internal/Dockerfile-with-features with content: 
 # syntax=docker.io/docker/dockerfile:1.4
ARG _DEV_CONTAINERS_BASE_IMAGE=placeholder
FROM  AS final

FROM $_DEV_CONTAINERS_BASE_IMAGE AS dev_containers_target_stage

USER root

COPY ./.devpod-internal/ /tmp/build-features/
RUN chmod -R 0755 /tmp/build-features && ls /tmp/build-features

RUN \\
echo \"_CONTAINER_USER_HOME=$(getent passwd root | cut -d: -f6)\" >> /tmp/build-features/devcontainer-features.builtin.env && \\
echo \"_REMOTE_USER_HOME=$(getent passwd vscode | cut -d: -f6)\" >> /tmp/build-features/devcontainer-features.builtin.env

ENV NVM_DIR=/usr/local/share/nvm
ENV NVM_SYMLINK_CURRENT=true
ENV PATH=/usr/local/share/nvm/current/bin:${PATH}
RUN cd /tmp/build-features/0 \\
&& chmod +x ./devcontainer-features-install.sh \\
&& ./devcontainer-features-install.sh



ARG _DEV_CONTAINERS_IMAGE_USER=root
USER $_DEV_CONTAINERS_IMAGE_USER","level":"debug"}}
[...]

PR #1530, opened 2025-01-09
Feature/fix compose multi stage builds

bkneis:
Fixes - https://github.com/loft-sh/devpod/issues/1433

The issue was that during our build pipeline we assumed an Image was used instead of a dockerfile, resulting in a `FROM empty string AS final` being added. The empty string came from the compose's devcontainer image name, which is nil since we are building the devcontainer.

The fix was to assign the dockerFileContents to the read docker file if no modifications were made. Looking at `pkg/devcontainer/parse.go` line 223 we perform a check on the target and potentially return an empty string, as the dockerfile does not need to be modified. In this case we need to simply assign dockerFileContents to the variable containing the original dockerfile (line 450 in `pkg/devcontainer/compose.go`).

I also added 2 examples, one for multi stage builds and a compose. You can use these to validate the fix by seeing the OP's error running `devpod up examples/compose` then running the same with this PR's build.


.env
```diff
@@ -0,0 +1,2 @@
+TEST=test
+COMPOSE_PROJECT_NAME=test
```
janekbaraniewski, line 2: do we still need this file in project root? looks like a leftover from ./examples/compose

examples/build-multi-stage/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+    apt-get update \
+    && apt-get install -y --no-install-recommends protobuf-compiler \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+FROM go AS final
+
+COPY app /app
+
+RUN echo hello
```
janekbaraniewski, line 31: new line 😅

examples/compose/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    apt-get update \
+    && apt-get install -y --no-install-recommends protobuf-compiler \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+FROM go AS final
+
+COPY app /app
+
+RUN echo hello
```
janekbaraniewski, line 32: new line 😅

janekbaraniewski (approved):
Awesome, just few nitpicks 😄  really like what you did with including more comments in code handling compose configuration!

I was thinking maybe we could also extract some of those pieces of code to smaller functions (like `readOriginalDockerfile` etc) so that this `buildAndExtendDockerCompose` function becomes a bit smaller and easier to grasp and also self-documents better. wdyt?

.env
bkneis: whoops, good catch yes that was from testing

bkneis:
> Awesome, just few nitpicks 😄 really like what you did with including more comments in code handling compose configuration!
> 
> I was thinking maybe we could also extract some of those pieces of code to smaller functions (like `readOriginalDockerfile` etc) so that this `buildAndExtendDockerCompose` function becomes a bit smaller and easier to grasp and also self-documents better. wdyt?

yes for sure that sounds good, I'll try that now

Issue #1537, opened 2025-01-10
DevPod should send the desired log level to providers


mrsimonemms:
**Is your feature request related to a problem?**  
I'm a maintainer of a [provider](https://github.com/mrsimonemms/devpod-provider-hetzner) and I log things at different levels. By default, the logger is set to `Info`. If the DevPod application is in debug mode, there should be a way of the application telling the provider to emit messages at the `Debug` level.

Often, the debug level has useful information for debugging problems (who knew?), but that we don't want to see in normal circumstances.

**Which solution do you suggest?**  
All the communication between the application and the provider is done via [environment variables](https://github.com/loft-sh/devpod/blob/main/pkg/provider/env.go) - it would make sense to use that mechanism. My suggestions are either:

- `DEBUG`: `true|false`
- `LOG_LEVEL`: `debug|info|warn` (this would default to `info` if nothing sent)

It looks like the application only recognises whether it's in debug mode or not, so the top one would make sense. But having a `LOG_LEVEL` would allow for granularity and future changes.

In truth, I don't really care which is which, so long as something is set.

**Which alternative solutions exist?**  
Do nothing - not a major problem, but does miss out of the ability to see debug messages

**Additional context**
[Slack thread](https://loft-sh.slack.com/archives/C056ZDZPJ4W/p1736525243011999)

PR #1540, opened 2025-01-11
feat: add log level to environment variables sent to provider

mrsimonemms:
## Description
This adds a `DEVPOD_LOG_LEVEL` environment variable to the environment variables sent to the provider.

### Design decisions

- I chose this approach because it allows for further granular log levels in future.
- I used the `DEVPOD_` prefix because it seemed to be in that section of options.

### Discussion

Any data received from the provider will be output at the `info` level because it's only doing a simple "send everything to `info`" rather than receiving any metadata to say the log level. My opinion would be to get this in today to allow debug logs to be output and then overhaul the log metadata as a future task.

## Issue
Fixes https://github.com/loft-sh/devpod/issues/1537

## How to test

In order to receive the log level in the provider, you should do something like [this](https://github.com/mrsimonemms/devpod-provider-hetzner/blob/6465c90933e72a6259c0483dd3917933577ba4e4/cmd/root.go#L32,L45).

Now run the binary:

```bash
devpod up --debug <target>
```

This will then tell the provider to set to the log level to `debug`.

### DIFF
--- a/pkg/provider/env.go
+++ b/pkg/provider/env.go
@@ -8,13 +8,15 @@ import (
 	"strings"
 
 	"github.com/loft-sh/devpod/pkg/config"
+	log2 "github.com/loft-sh/log"
 )
 
 const (
 	// general
-	DEVPOD      = "DEVPOD"
-	DEVPOD_OS   = "DEVPOD_OS"
-	DEVPOD_ARCH = "DEVPOD_ARCH"
+	DEVPOD           = "DEVPOD"
+	DEVPOD_OS        = "DEVPOD_OS"
+	DEVPOD_ARCH      = "DEVPOD_ARCH"
+	DEVPOD_LOG_LEVEL = "DEVPOD_LOG_LEVEL"
 
 	// workspace
 	WORKSPACE_ID       = "WORKSPACE_ID"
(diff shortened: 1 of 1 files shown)


pascalbreuninger (approved):
LGTM, thanks!

mrsimonemms:
No problem @pascalbreuninger

I'm not 100% certain, but I don't think the failing test is any change I've made but looks like a Docker failure.

mrsimonemms:
@pascalbreuninger I was just starting to go through the providers to add the `DEVPOD_LOG_LEVEL` envvars into each one and found `DEVPOD_DEBUG` in the [Kubernetes provider](https://github.com/loft-sh/devpod-provider-kubernetes/blob/main/cmd/root.go#L22)

Having a look through, it looks like this [actually a thing](https://github.com/loft-sh/devpod/blob/b35dbf32bd08101fd7cc5d775cbdf8e5bd310348/pkg/client/clientimplementation/proxy_client.go#L26).

Question - do you want to keep both `DEVPOD_DEBUG` and `DEVPOD_LOG_LEVEL`? It doesn't feel a great idea to maintain two ways of doing the same thing.

Unless you think you'll change to having different log levels, I would suggest reverting this PR. Once you've done that, I'll update all the official providers to use `DEVPOD_DEBUG` (and my Hetzner provider). I'll also add a line to the provider docs too.

PR #1575, opened 2025-01-18
fix(Zed): remove the double leading slashes on project

acolombier:
This fixes an issue where project opened in Zed are opened as `//workspace/project/` instead of `/workspace/project/`. generally, the double slashes doesn't seem to prevent Zed to function. However, a few alert may arise in the logs depending of the used LSP and extensions.


pascalbreuninger (commented):
@acolombier LGTM, thanks for your contribution! I wasn't even aware of the problem

pascalbreuninger (approved):

Issue #1573, opened 2025-01-18
Devpod UI not showing up


Rohamgames:
**What happened?**  
Just installed Devpod on a Windows machine and I can see the Devpod icon in the status bar, but nothing shows up even when I click show dashboard.
Tried reinstalling a few times nothing chnaged.

**What did you expect to happen instead?**  
Just like my machine it should launch as well. (the other running Windows 11 and this one running LTSC version)

**How can we reproduce the bug?** (as minimally and precisely as possible)  
I dont know, but I'm willing to send logs (unfortunately I don't know where they are either) 

I havn't reached the container part yet.

My *`devcontainer.json`*:
```
{
    "name": "...",
    ...
}
```

**Local Environment:**  
- DevPod Version: 0.6.9
- Operating System: Windows 11 LTSC
- ARCH of the OS: AMD64 

**Anything else we need to know?**  
I remember that I have stopped some of the windows services in the past on this machine to save some resources so that may be a reason, what services Devpod requries to be able to show me that UI are there any dependencies?
All the other apps and softwares are working as expected
Also it is not an anti virus thing to be clear.

Rohamgames:
Looks like #1249

Rohamgames:
I have tried the same procedure on Windows Sandbox and it is working just fine with no issues
 
![Image](https://github.com/user-attachments/assets/67419315-d75f-4dec-ac60-f17f2bc41da1)

Rohamgames:
I have tried installing the app using winget and the issue is still there, no UI shows up, but the process is visible in taskmanager

![Image](https://github.com/user-attachments/assets/da1dc485-c1e8-47d1-a80d-83f56ec86d1e)

Rohamgames:
Well, I have figured out it is not showing UI due to [WebView2 ](https://developer.microsoft.com/en-us/microsoft-edge/webview2/) not being present

It would be nice if we could add some error dialog to notify the client that they are missing the required component

bkneis:
Thanks for reporting @Rohamgames! I've updated the docs now so anyone facing the same issue can resolve

PR #1582, opened 2025-01-20
Update docs for windows dependency

bkneis:
OSS user discovered windows install failed due to missing dependency - https://github.com/loft-sh/devpod/issues/1573

This PR updates the docs to inform users of the required package

### DIFF
--- a/docs/pages/getting-started/install.mdx
+++ b/docs/pages/getting-started/install.mdx
@@ -43,6 +43,11 @@ Make sure you have the following dependencies installed for the Appimage to work
 Note that these are generally needed for AppImage to work, they are not specific to DevPod
 :::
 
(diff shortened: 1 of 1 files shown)


pascalbreuninger:
WebView2 should ship with most modern Windows Installations (10 and 11). Could you add that as a note for older windows installations please? 
I'd like to avoid confusion where people go and look for WebView because for 99% it should already exist

bkneis:
> WebView2 should ship with most modern Windows Installations (10 and 11). Could you add that as a note for older windows installations please? I'd like to avoid confusion where people go and look for WebView because for 99% it should already exist

Done :)

pascalbreuninger (approved):
LGTM

Issue #1554, opened 2025-01-14
"runServices" not being used when re-running workspace


iJackUA:
**What happened?**  
When I do subsequent (after initial `up` or recreate) runs of the workspace - it ignores `runServices` and starts ALL containers from docker-compose.yml, not only those specified.
If I do `devpod up ... --recreate` - it starts services as expected (as first start). But each next stop/up cycle - starts all unneeded containers.

**What did you expect to happen instead?**  
On any `up` of the workspace to start only containers specified in `runServices`

**How can we reproduce the bug?** (as minimally and precisely as possible)  

I have minimalistic repo with DevContainer and DockerCopomose where issue can be seen
https://github.com/iJackUA/gitpod-devcontainer-test/tree/main

1) Initial workspace up starts as expected: `web`, `db`
`devpod up git@github.com:iJackUA/gitpod-devcontainer-test.git@main --id devpod-test --ide vscode`

2) Stop: stops all containers as expected
`devpod stop devpod-test`

3) Run workspace again. It does UNEXPECTED, starts: `web`, `db`, `sidecar1`, `sidecar2`
`devpod up devpod-test`

Attach logs with `--debug` per each step below

My *`devcontainer.json`*:
```
{
  "name": "Ubuntu",
  "dockerComposeFile": ["../docker-compose.yml", "docker-compose.yml"],
  "service": "web",
  "runServices": ["web", "db"],
  "workspaceFolder": "/app"
}
```

**Local Environment:**  
- DevPod Version: v0.6.8
- Operating System: mac
- ARCH of the OS: ARM64

**DevPod Provider:**  
- Cloud Provider: -
- Kubernetes Provider: -
- Local/remote provider: ssh
- Custom provider: -

**Anything else we need to know?**  

1) `devpod up git@github.com:iJackUA/gitpod-devcontainer-test.git@main --id devpod-test --ide vscode`

```
11:42:53 debug No pro instance available, skipping provider upgrade check
11:42:53 debug Acquire workspace lock...
11:42:53 debug Acquired workspace lock...
11:42:53 info Creating devcontainer...
11:42:53 debug Adding ssh keys to agent, disable via 'devpod context set-options -o SSH_ADD_PRIVATE_KEYS=false'
11:42:53 debug Inject and run command: '/tmp/ijackua/devpod/agent' helper ssh-server --stdio --debug
11:42:53 debug execute inject script
[...]

PR #1583, opened 2025-01-21
Refactor runServices upArgs logic

bkneis:
Fixes https://github.com/loft-sh/devpod/issues/1554

The issue was that the initial up logic that ran docker-compose iterated through `runServices` and only added those to the upArgs (line 372). The condition when the container already existed (a re run of up) did not do this (line 167). The fix was to add the same logic to add to upArgs only those services defined by runServices. Additionally I refactored it to a function

### DIFF
--- a/pkg/devcontainer/compose.go
+++ b/pkg/devcontainer/compose.go
@@ -211,9 +211,9 @@ func (r *runner) runDockerCompose(
 
 // onlyRunServices appends the services defined in .devcontainer.json runServices to the upArgs
 func (r *runner) onlyRunServices(upArgs []string, parsedConfig *config.SubstitutedConfig) []string {
+	// Always run the main devcontainer
+	upArgs = append(upArgs, parsedConfig.Config.Service)
 	if len(parsedConfig.Config.RunServices) > 0 {
-		// Run the main devcontainer
-		upArgs = append(upArgs, parsedConfig.Config.Service)
 		// Run the services defined in .devcontainer.json runServices
 		for _, service := range parsedConfig.Config.RunServices {
 			if service == parsedConfig.Config.Service {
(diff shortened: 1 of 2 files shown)


pkg/devcontainer/compose.go
```diff
@@ -206,6 +209,22 @@ func (r *runner) runDockerCompose(
 	return r.setupContainer(ctx, parsedConfig.Raw, containerDetails, mergedConfig, substitutionContext, timeout)
 }
 
+// onlyRunServices appends the services defined in .devcontainer.json runServices to the upArgs
+func (r *runner) onlyRunServices(upArgs []string, parsedConfig *config.SubstitutedConfig) []string {
+	if len(parsedConfig.Config.RunServices) > 0 {
+		// Run the main devcontainer
+		upArgs = append(upArgs, parsedConfig.Config.Service)
```
pascalbreuninger, line 216: Don't we still need this if `len(runServices) == 0`?

pkg/devcontainer/compose.go
```diff
@@ -206,6 +209,22 @@ func (r *runner) runDockerCompose(
...
+		// Run the services defined in .devcontainer.json runServices
+		for _, service := range parsedConfig.Config.RunServices {
```
pascalbreuninger, line 218: You might be able to get rid of the `if` statement above if we always need to append `parsedConfig.Config.Service` because `range` handles `nil` slices properly

pkg/devcontainer/compose.go
bkneis: Yes I was thinking that, but didn't want to change the existing behaviour. I wasn't sure if runServices was omitted if we wanted all services to run alongside the devcontainer?

bkneis: I couldn't see anything in the spec that said this though, so maybe we should assume we should always run it? even without runServices

pascalbreuninger (approved):
LGTM

PR #1589, opened 2025-01-23
Feature/tailscale network tools

bkneis:
This PR contains the addition of the following CLI commands

`devpod ts ping`, `devpod ts netcheck`, `devpod ts status` and `devpod ts metrics`. Each are the same as running `tailscale` instead of `devpod ts` in the CLI. They were taken from https://github.com/tailscale/tailscale/tree/v1.78.3/cmd/tailscale/cli where the CLI tool was ported to cobra and the tailscale `LocalClient` was injected by starting a tailscale server using `pkg/tailscale` and returning it's userspace local client.

One open question in this PR are my changes to `tsnet.go`. I changed the call to `Start` with `Up`, [looking at the docs](https://pkg.go.dev/tailscale.com/tsnet#Server.Up) Start and Up are the same except Up waits for the client to be connected. I thought this would be preferably than having to poll a local tailscale endpoint to check it. I then passed a channel to signal when the call to `Start` had completed calling `Up` (therefore was connected to the network) and before blocking on the context. Let me know your thoughts


cmd/ts/tailscale.go
```diff
@@ -0,0 +1,34 @@
...
+	"strings"
+
+	"github.com/loft-sh/devpod/cmd/agent"
+	"github.com/loft-sh/devpod/cmd/flags"
+	"github.com/spf13/cobra"
+)
+
+func NewTailscaleCmd(globalFlags *flags.GlobalFlags) *cobra.Command {
+	tsCmd := &cobra.Command{
+		Use:   "ts",
```
pascalbreuninger, line 16: Can you make that command `hidden` please?

pkg/tailscale/tsnet.go
```diff
@@ -81,7 +83,7 @@ func (t *tsNet) Start(ctx context.Context) error {
 	}
 
 	// Start the server
-	if err := t.tsServer.Start(); err != nil {
+	if _, err := t.tsServer.Up(ctx); err != nil {
```
pascalbreuninger, line 86: LGTM, don't see a reason not to do it that way 👍

cmd/ts/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+	// Create network
+	tsNet := tailscale.NewTSNet(&tailscale.TSNetConfig{
+		AccessKey: cmd.AccessKey,
+		Host:      tailscale.RemoveProtocol(cmd.PlatformHost),
+		Hostname:  cmd.NetworkHostname,
+	})
+	// Run tailscale up and wait until we have a connected client
+	done := make(chan bool)
+	go func() {
+		err := tsNet.Start(ctx, done)
```
pascalbreuninger, line 57: Aren't we in a deadlock if `tsNet.Start` errors out here since we never send a message to `done`?

cmd/ts/netcheck.go
```diff
@@ -0,0 +1,269 @@
...
+	// Create network
+	tsNet := tailscale.NewTSNet(&tailscale.TSNetConfig{
+		AccessKey: cmd.AccessKey,
+		Host:      tailscale.RemoveProtocol(cmd.PlatformHost),
+		Hostname:  cmd.NetworkHostname,
+	})
+	// Run tailscale up and wait until we have a connected client
+	done := make(chan bool)
+	go func() {
+		err := tsNet.Start(ctx, done)
```
pascalbreuninger, line 62: Same as in `metrics.go`. This applies to all commands, won't add individual comments for each one

pascalbreuninger (commented):
LGTM overall, a couple of smaller comments here and there.

Bigger topic: What do you think of moving all logging to our logger implementation?
Right now it always prints to Stdout/err which will make it harder for us in the future to expose these commands from within a workspace.

cmd/ts/netcheck.go
```diff
@@ -0,0 +1,269 @@
...
+	netcheckCmd.Flags().StringVar(&cmd.format, "format", "", `output format; empty (for human-readable), "json" or "json-line"`)
+	netcheckCmd.Flags().DurationVar(&cmd.every, "every", 0, "if non-zero, do an incremental report with the given frequency")
+	netcheckCmd.Flags().BoolVar(&cmd.verbose, "verbose", false, "verbose logs")
+
+	cmd.ParseFlags(netcheckCmd)
+	return netcheckCmd
+}
+
+func (cmd *NetcheckCmd) Run(_ *cobra.Command, _ []string) error {
+	ctx := context.Background()
```
janekbaraniewski, line 52: wdyt about using:
```
ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
``` 
?

cmd/ts/netcheck.go
```diff
@@ -0,0 +1,269 @@
...
+			if cmd.verbose {
+				derpNum = fmt.Sprintf("derp%d, ", rid)
+			}
+			printf("\t\t- %3s: %-7s (%s%s)\n", r.RegionCode, latency, derpNum, r.RegionName)
+		}
+	}
+	return nil
+}
+
+func portMapping(r *netcheck.Report) string {
```
janekbaraniewski, line 229: can we move some of this stuff from `cmd/` to `pkg/tailscale`?

cmd/ts/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+func (flags *TsNetFlags) ParseFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&flags.AccessKey, "access-key", "", "")
+	cmd.Flags().StringVar(&flags.PlatformHost, "host", "", "")
+	cmd.Flags().StringVar(&flags.NetworkHostname, "hostname", "", "")
+}
+
+func NewMetricsCmd() *cobra.Command {
+	cmd := &MetricsCmd{&TsNetFlags{}}
+	var metricsCmd = &cobra.Command{
+		Use:   "metrics print",
```
janekbaraniewski, line 32: can we use a single string with no white spaces?

cmd/ts/ping.go
```diff
@@ -0,0 +1,244 @@
...
+	icmp        bool
+	peerAPI     bool
+	timeout     time.Duration
+}
+
+func NewPingCmd() *cobra.Command {
+	cmd := &PingCmd{TsNetFlags: &TsNetFlags{}}
+	var pingCmd = &cobra.Command{
+		Use:     "ping",
+		Example: "tailscale ping <hostname-or-IP>",
```
janekbaraniewski, line 39: this is based on open source tailscale client implementation right? Can we put links to original files somewhere at the top?

cmd/ts/ping.go
```diff
@@ -0,0 +1,244 @@
...
+	if cmd.icmp {
+		return tailcfg.PingICMP
+	}
+	if cmd.peerAPI {
+		return tailcfg.PingPeerAPI
+	}
+	return tailcfg.PingDisco
+}
+
+func (cmd *PingCmd) runPing(ctx context.Context, args []string, localClient *ts.LocalClient) error {
```
janekbaraniewski, line 112: can we also move all those funcs to `pkg/tailscale`?

cmd/ts/status.go
```diff
@@ -0,0 +1,355 @@
...
+
+func NewStatusCmd() *cobra.Command {
+	cmd := &StatusCmd{TsNetFlags: &TsNetFlags{}}
+
+	var statusCmd = &cobra.Command{
+		Use:     "status",
+		Example: "tailscale status [--active] [--web] [--json]",
+		Short:   "Show state of tailscaled and its connections",
+		Args:    cobra.NoArgs,
+		RunE: func(cobraCmd *cobra.Command, args []string) error {
```
janekbaraniewski, line 49: can we move it to `cmd.Run`?

cmd/ts/status.go
```diff
@@ -0,0 +1,355 @@
...
+		go func() {
+			<-ctx.Done()
+			ln.Close()
+		}()
+		if cmd.browser {
+			go open2.Open(ctx, statusURL, nil)
+		}
+		err = http.Serve(ln, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if r.RequestURI != "/" {
+				http.NotFound(w, r)
```
janekbaraniewski, line 133: this is definitely way too many nested if statements 😅 can this be somehow refactored into smaller functions?

pkg/tailscale/tsnet.go
janekbaraniewski: yeah this is way better, we can get network info from Up result and put it in WorkspaceInstance status if we ever need to 👍

janekbaraniewski (approved):
LGTM! Couple comments about formatting etc

pkg/tailscale/tsnet.go
bkneis: :+1:

cmd/ts/netcheck.go
bkneis: neat! Didn't know that existed. yes sure

cmd/ts/metrics.go
bkneis: I could, I was trying to mirror the tailscale CLI as is so that you could literally alias `tailscale` for `devpod ts` so the tailscale docs remain the same. Also we then have the potential to steal other scripts etc in the future if they expect the tailscale CLI. What do you think? https://tailscale.com/kb/1080/cli#metrics. Happy either way though as I can see advantages to us changing the code for our specific debugging tasks as well.

cmd/ts/ping.go
bkneis: yup, I added a docblock to the package but I'll add the exact file at the top of each of the commands :+1:

cmd/ts/metrics.go
bkneis: No because we run `log.Fatalf` in the event of an error which calls `os.Exit(1)`

bkneis:
@pascalbreuninger good point, I guess this depends on how we intend to use the tool. So far I have been running in an SSH session on my workspace. I suppose the logger is only needed if we execute the commands in the SSH tunnel pragmatically? Would we do this? If not logging to STDOUT should suffice. Let me know if this needs to be implemented but I thought this would be run manually by the user

cmd/ts/netcheck.go
bkneis: I could do, are there functions here you think we could use elsewhere? The rationale here was that if we update our tailscale fork we will need to sync these commands with the upstream. So I thought the least refactoring I did the easier the updates would be. For instance most of these files are almost the exact same except for the first 50 lines, which are replaced by the cobra struct and injecting of the local client.

But if there are useful functions in here and you think it is worth us having this be more structured then happy to refactor like you said and move stuff to `pkg/tailscale` and clean up some of the harder to read code

cmd/ts/metrics.go
janekbaraniewski: sorry forgot to check this PR 🤦 yeah sounds good, if it's mirroring upstream then lets change as little as possible

PR #1621, opened 2025-02-07
feat: Add display of "Created" and "Last activity" dates to list of p…

PRTTMPRPHT:
…ro workspaces


desktop/src/views/Pro/Workspace/WorkspaceInstanceCard.tsx
```diff
@@ -167,6 +185,18 @@ export function WorkspaceInstanceCard({ instanceName, host }: TWorkspaceInstance
               </Text>
             </WorkspaceInfoDetail>
 
+            {lastActivity && (
+              <WorkspaceInfoDetail icon={Clock} label={<Text>Last activity</Text>}>
```
pascalbreuninger, line 189: Could we show the full timestamp as a tooltip here and in the `created` section?

pascalbreuninger (approved):

PR #1644, opened 2025-02-14
Docs(insatllation): requires libwebkit2gtk-4.1

robsdedude:
Otherwise I was getting this error on trying to start DevPod

```
DevPod: error while loading shared libraries: libwebkit2gtk-4.1.so.0: cannot open shared object file: No such file or directory
```


pascalbreuninger (approved):
LGTM, thanks

PR #1677, opened 2025-02-27
Add flatpak to release assets

bkneis:
This PR adds devpod's app metadata so it can be pulled by flathub. This was requested in my PR to flathub to publish our flatpak, currently it is in the flathub repo itself. 

A templated flatpak manifest has also been added so that we can automatically update the version and hash of the devpod binary. Once this PR is merged, I can update my flathub PR and once approved, devpod will be available on flathub with it's own git repo. When the repo is available I can submit another PR for the CI to push our manifest to it when we cut a new release to update devpod is flathub.

For more info on flathub releases see

https://docs.flathub.org/docs/for-app-authors/submission#after-submission

https://docs.flathub.org/docs/for-app-authors/maintenance


.github/workflows/release.yaml
```diff
@@ -448,6 +448,73 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Upload flatpak assets
+        if: matrix.settings.host != 'windows-latest'
```
pascalbreuninger, line 452: Can you replace this with `        if: matrix.settings.host == 'ubuntu-22.04' && matrix.settings.cli_only == false` please? Otherwise it'll also run for macos and cli only builds

.github/workflows/release.yaml
```diff
@@ -448,6 +448,73 @@ jobs:
...
+              name: mdAssetName,
+              data: fs.readFileSync(mdAssetPath),
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              release_id: releaseId
+            })
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Generate updated flatpak manifest
```
pascalbreuninger, line 495: Can you put the same `if` as above for this and the following step so that we don't run in for windows/macos please?

pascalbreuninger (changes requested):

pascalbreuninger (approved):
LGTM

PR #1691, opened 2025-03-01
refactor: use task based runner

FabianKramm:


pkg/platform/env.go
```diff
@@ -11,7 +11,7 @@ const (
 
 	ConfigEnv  = "LOFT_CONFIG"
 	ProjectEnv = "LOFT_PROJECT"
-	RunnerEnv  = "LOFT_RUNNER"
+	ClusterEnv = "LOFT_CLUSTER"
```
pascalbreuninger, line 14: We'd still need the `RunnerEnv` constant for older clients

pascalbreuninger (approved):

PR #1713, opened 2025-03-05
Run devpod daemon as root process in pro workspaces

janekbaraniewski:


cmd/agent/container/daemon.go
```diff
@@ -1,72 +1,215 @@
...
 	}
 
-	err = os.WriteFile(agent.ContainerActivityFile, nil, 0o777)
+	// Listen for OS signals.
+	go handleSignals(ctx, errChan)
+
+	// Wait until an error (or termination signal) occurs.
+	err := <-errChan
+	cancel()
+	wg.Wait()
```
pascalbreuninger, line 100: Do we still need the waitgroup if we have the `errChan`?

cmd/agent/container/daemon.go
```diff
@@ -1,72 +1,215 @@
...
+
+	// Use the error value before exiting.
 	if err != nil {
-		return err
+		fmt.Fprintf(os.Stderr, "Daemon error: %v\n", err) // Logging the error for visibility [1]
```
pascalbreuninger, line 104: 👍

pkg/devcontainer/single.go
```diff
@@ -81,11 +87,13 @@ func (r *runner) runSingleContainer(
 			}
 		}
 	} else {
+
 		// we need to build the container
 		buildInfo, err := r.build(ctx, parsedConfig, substitutionContext, provider2.BuildOptions{
 			CLIOptions: provider2.CLIOptions{
 				PrebuildRepositories: options.PrebuildRepositories,
 				ForceDockerless:      options.ForceDockerless,
+				Platform:             options.Platform,
```
pascalbreuninger, line 96: Is this required?

cmd/agent/container/daemon.go
janekbaraniewski: yeah we could read from errChan from either routine so in case both are running we need a waitgroup. Question is if that can ever happen 😅

pkg/devcontainer/single.go
janekbaraniewski: ah thx! missed it while cleaning up

janekbaraniewski: updated 👍

pascalbreuninger (approved):

PR #1749, opened 2025-03-12
Check kube config for the namespace before using the default

bkneis:
When the kubernetes driver was moved into devpod, a customer discovered an issue where the PVC is always scheduled in the default namespace "devpod". This could be reproduced by deploying any workspace using a space template.

This PR fixes the issue by first checking the kube config, and if available use this namespace, instead of the default.

I tested this against a local platform with the only change being this build, using a space template. I also validated the namespace being used in the logs


pascalbreuninger (approved):
LGTM

PR #1770, opened 2025-03-17
Disable auto updates for linux

bkneis:
This PR disables the auto updater window for linux since updates will be managed through our flathub repo


bkneis:
Superseded by https://github.com/loft-sh/devpod/pull/1826

PR #1779, opened 2025-03-19
CLI usability fixes

janekbaraniewski:


pkg/workspace/list.go
```diff
@@ -185,9 +185,9 @@ func listProWorkspacesForProvider(ctx context.Context, devPodConfig *config.Conf
 	}
 	if err != nil {
 		if log.GetLevel() < logrus.DebugLevel {
-			log.Warnf("Failed to list pro workspaces for provider %s", provider)
+			log.Warnf("Failed to list pro workspaces for provider %s. Is DevPod daemon running?", provider)
```
pascalbreuninger, line 188: That's only relevant for the daemon provider though 🤔

PR #1806, opened 2025-03-28
Update tauri typescript and rust to 2.4.0 and plugins

bkneis:


desktop/src-tauri/src/main.rs
```diff
@@ -93,6 +93,7 @@ fn main() -> anyhow::Result<()> {
             resources_handles: Arc::new(Mutex::new(vec![])),
         })
         .plugin(logging::build_plugin())
+        // .plugin(tauri_plugin_deep_link::init())
```
FabianKramm, line 96: Is that wanted? That will basically disable the deep links doesn't it?

bkneis: oops, I should remove that. No i didn't end up updating that plugin as it didn't work. That was my attempt to use the new API that had a breaking change

PR #1826, opened 2025-04-07
Feature/flatpak

bkneis:
This PR makes the application changes required to run devpod in a flatpak sandbox. We now pass the env var `DEVPOD_ADDITIONAL_ENV_VARS` through to propagate environment variables from the sandbox to host. Additionally the install CLI feature is fixed and the front end can now call rust to know if it is running as a flatpak.

Once this is merged I can reference the build in the flathub PR - https://github.com/flathub/flathub/pull/6223 and everything should be good to go :)


desktop/src-tauri/src/install_cli.rs
```diff
@@ -92,6 +94,9 @@ fn install(_app_handle: AppHandle, force: bool) -> Result<(), InstallCLIError> {
         let mut user_local_bin = home;
         user_local_bin.push(".local/bin/devpod");
 
+        // create .local/bin if necessary
+        fs::create_dir_all(user_local_bin.clone().parent().unwrap());
```
pascalbreuninger, line 98: This panics if it can't create the parent dir, is that expected?

desktop/src/client/command.ts
```diff
@@ -57,30 +61,37 @@ export class Command implements TCommand<ChildProcess<string>> {
...
   public async run(): Promise<Result<ChildProcess<string>>> {
     try {
+      // Run once to check with the rust backend if we are running inside the flatpak sandbox
+      // This informs the CLI wrapper to use flatpak-spawn to escape the sandbox and export this.extraEnvVars
+      if (this.isFlatpak === undefined) {
+        this.isFlatpak = await this.getEnv("FLATPAK_ID")
+        if (this.isFlatpak) {
+          this.extraEnvVars["FLATPAK_ID"] = "sh.loft.devpod"
+          this.extraEnvVars[DEVPOD_ADDITIONAL_ENV_VARS] = recordToCSV(this.extraEnvVars)
+          this.sidecarCommand = ShellCommand.sidecar(DEVPOD_BINARY, this.args, { env: this.extraEnvVars })
```
pascalbreuninger, line 92: Doesn't this need to execute the `run-path-devpod-wrapper` anymore?

bkneis: No, as far as I can tell this was a half baked solution. `run-path-devpod-wrapper` doesn't exist, `/app/bin/devpod-cli` which is `DEVPOD_BINARY` when in flatpak, is already pointing to the shim because of https://github.com/flathub/flathub/pull/6223/files#diff-e8ce64f25e7da6a0311004091b01bfdf3651499f91a25f5c8cce57a6a434aa4eR42

desktop/src-tauri/src/install_cli.rs
bkneis: Updated, I added `.unwrap()` which I believe is what's needed?

pascalbreuninger (approved):

PR #1836, opened 2025-04-09
WIP: Rework git credentials helper server

janekbaraniewski:


pkg/daemon/local/credentials_server.go
```diff
@@ -0,0 +1,129 @@
...
+	"net/http"
+	"time"
+
+	"github.com/loft-sh/devpod/pkg/gitcredentials"
+	"github.com/loft-sh/log"
+	"tailscale.com/tsnet"
+)
+
+const (
+	LocalCredentialsServerPort = 9999
```
pascalbreuninger, line 18: Can we use a random port here please?

pkg/tunnel/services.go
```diff
@@ -109,14 +121,33 @@ func RunServices(
 			close(errChan)
 		}()
 
-		// run credentials server
+		log.Infof("Waiting for credentials server port to be assigned...")
```
pascalbreuninger, line 124: Nit: can you make this `debug` so that it doesn't show up when just `ssh`ing please?

pkg/tunnel/services.go
```diff
@@ -109,14 +121,33 @@ func RunServices(
...
+		var port int
+		select {
+		case port = <-portChan:
+			log.Infof("Credentials server running on port %d\n", port)
```
pascalbreuninger, line 128: This logs as `Credentials server running on port 0` every time when not used with pro, can we only print that if it's != 0?

janekbaraniewski: updadet 👍

pkg/tunnel/services.go
janekbaraniewski: updated 👍

deniseschannon (approved):

PR #1884, opened 2025-06-18
Add support for platform user docker credentials

janekbaraniewski:


cmd/agent/docker_credentials.go
```diff
@@ -146,3 +161,64 @@ func (cmd *DockerCredentialsCmd) handleGet(log log.Logger) error {
...
+			DialContext: func(_ context.Context, _, _ string) (net.Conn, error) {
+				return net.Dial("unix", filepath.Join(container.RootDir, ts.RunnerProxySocket))
+			},
+		},
+	}
+
+	credentials, credentialsErr := requestDockerCredentials(httpClient, credentials, "http://runner-proxy/docker-credentials")
+	if credentialsErr != nil {
+		// append error to /tmp/docker-credentials.log
+		file, err := os.OpenFile("/tmp/docker-credentials-error.log", os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
```
pascalbreuninger: We might want put this into `/var/devpod` because that's guaranteed to exist and be writable at this point.

cmd/agent/docker_credentials.go
```diff
@@ -146,3 +161,64 @@ func (cmd *DockerCredentialsCmd) handleGet(log log.Logger) error {
...
+
+func getDockerCredentialsFromWorkspaceServer(credentials *dockercredentials.Credentials) *dockercredentials.Credentials {
+	if _, err := os.Stat(filepath.Join(container.RootDir, ts.RunnerProxySocket)); err != nil {
+		// workspace server is not running
+		return nil
+	}
+
+	httpClient := &http.Client{
+		Transport: &http.Transport{
+			DialContext: func(_ context.Context, _, _ string) (net.Conn, error) {
```
pascalbreuninger, line 174: Is there a default timeout? If not, might want to limit it to 15 seconds or so

janekbaraniewski: yeah good point, added 15s timeout to client

cmd/agent/docker_credentials.go
janekbaraniewski: 👍 moved

pascalbreuninger (approved):
LGTM

PR #1944, opened 2025-11-26
feat: Add SSH_CONFIG_INCLUDE_PATH option for read-only SSH configs

clord:
## Summary
Adds support for read-only SSH config files (e.g., managed by Nix Home Manager.)
New context option `SSH_CONFIG_INCLUDE_PATH` specifies an alternate file where DevPod
writes host entries. Then the main SSH config can use an `Include` directive to load 
DevPod-managed entries

## Motivation
Users with declaratively managed SSH configs (Nix, chezmoi, etc.) cannot have DevPod modify
`~/.ssh/config` directly since it's read-only or regenerated. This change allows DevPod to
write its host entries to a separate file that can be included.

## Usage

Add an Include directive to your SSH config file `~/.ssh/config`: 

```
   Include ~/.ssh/devpod_hosts
```

Configure DevPod to use the include path:

```bash
devpod context set-options -o SSH_CONFIG_INCLUDE_PATH=~/.ssh/devpod_hosts
```

DevPod will now write workspace host entries to `~/.ssh/devpod_hosts` instead of
`~/.ssh/config`.


skevetter:
If you send me your PR, I'd be to happy review/merge it https://github.com/skevetter/devpod (background context https://github.com/loft-sh/devpod/issues/1946).

PR #1971, opened 2026-03-20
Fix path separator issue on Windows

ykomatsu:
`filepath.Join` uses backslashes when running on Windows, but `workdir` is a path on the remote Linux host, which uses forward slashes.
I think that `path.Join` should be used here instead of `filepath.Join`.


YassineElbouchaibi:
Hello! I recommend raising on https://github.com/skevetter/devpod as the main repo is unmaintained.

